PR #1, opened 2016-03-07
Installing the test app from gradle

PrakashGoulla:
## Commit summary
This commit would let gradle install the app when `gradle assembleDebugAndroidTest` is run, on which the uiautomator2-server on device tests will be run.

I'm using selendroi-test-app as the test app, placed under `res/` directory.


PrakashGoulla:
@bootstraponline Please review

appium-uiautomator2-server/build.gradle
```diff
@@ -37,3 +37,17 @@ dependencies {
     androidTestCompile 'com.android.support.test.espresso:espresso-core:2.0'
 
 }
+
+task installAUT(type: Exec) {
+    def adb = android.getAdbExe().toString()
+    File configFile = file('../res/selendroid-test-app-debug.apk')
+    commandLine "$adb", 'install', '-r', "$configFile"
```
bootstraponline: `commandLine "$adb install -rg $apk".split(' ')`

I think it's cleaner to use split, we need the -g flag on api 23 and above, and config file isn't a good name for the apk variable.

appium-uiautomator2-server/build.gradle
```diff
@@ -37,3 +37,17 @@ dependencies {
     androidTestCompile 'com.android.support.test.espresso:espresso-core:2.0'
 
 }
+
+task installAUT(type: Exec) {
```
bootstraponline: needs to be like this:

```
task installAUT(type: Exec, dependsOn: 'installDebug') {
    group = 'test'
    description = 'Install app under test'
```

otherwise it won't get installed as part of the regular install debug flow.

appium-uiautomator2-server/build.gradle
```diff
@@ -37,3 +37,17 @@ dependencies {
...
+    def adb = android.getAdbExe().toString()
+    File configFile = file('../res/selendroid-test-app-debug.apk')
+    commandLine "$adb", 'install', '-r', "$configFile"
+}
+
+afterEvaluate {
+    tasks.each { task ->
+        if (task.name.startsWith('assembleDebugAndroidTest')) {
```
bootstraponline: It'd be good to preserve the comment.

`// Android Studio calls assemble directly not installDebug`

appium-uiautomator2-server/build.gradle
PrakashGoulla: after adding `dependsOn: 'installDebug'`, doing `gradle installDebug` is installing only the server apk but not the app under test.

PrakashGoulla: can it be because of this condition `if (task.name.startsWith('assembleDebugAndroidTest'))` under `afterEvaluate` ?

bootstraponline: If you run `gradle connectedCheck`, then it should install the server apk & AUT.

PrakashGoulla: `connectedCheck` installing both the apks, but hangs down while executing tests under `androidTest` which contains an infinite [loop](https://github.com/appium/appium-uiautomator2-server/blob/master/appium-uiautomator2-server/src/androidTest/java/io/appium/uiautomator2/AppiumUiAutomator2Server.java#L27) for keeping the server up.

PrakashGoulla: a workaround would be to add
```
 sourceSets {
        androidTest.setRoot('src/androidTest/tests')
    }
```

bootstraponline: I think we'd want a new test flavor. See [testFlavor1](https://github.com/googlesamples/android-testing-templates/tree/master/AndroidTestingBlueprint/app/src) from the android blueprint.

If the source set works though then that might be good enough for now.

PrakashGoulla:
updated

appium-uiautomator2-server/build.gradle
```diff
@@ -37,3 +42,21 @@ dependencies {
...
+    group = 'test'
+    description = 'Install app under test'
+    def adb = android.getAdbExe().toString()
+    def apk = file('../res/selendroid-test-app-debug.apk')
+    commandLine "$adb install -rg $apk".split(' ')
+}
+
+afterEvaluate {
+    // When launching individual tests from Android Studio, it seems that only the assemble tasks
+    // get called directly, not the install* versions
```
bootstraponline: if we're copying the wording, probably good to link to the original gist also so they don't complain later.

bootstraponline:
I made one comment, the code looks good to me.

The goal is to be able to run a `connectedCheck` which executes the tests on device.

PrakashGoulla:
The previous change was not installing the androidTest apk but only the src and AUT apks. updated to include `androidTest` apk as well.

I am working on another branch to include the tests that will be executed with `connectedCheck`.

bootstraponline:
> Does executing tests from Android Studio invokes connectedCheck?

Not sure but I think we want to be doing the install on `assembleDebugAndroidTest`, not on connectedCheck. Gradle has a lot of tasks that invoke assemble which may never call connectedCheck.

appium-uiautomator2-server/build.gradle
```diff
@@ -35,5 +40,25 @@ dependencies {
     androidTestCompile 'com.android.support.test.uiautomator:uiautomator-v18:2.1.2'
     androidTestCompile 'com.android.support.test:testing-support-lib:0.1'
     androidTestCompile 'com.android.support.test.espresso:espresso-core:2.0'
+    androidTestCompile 'com.squareup.okhttp:okhttp:2.5.0'
+
+}
 
+task installAUT(type: Exec, dependsOn: ['installDebug', 'installDebugAndroidTest']) {
```
bootstraponline: If `assembleDebugAndroidTest` is always invoked via afterEvaluate then maybe we can drop the dependsOn. You'll have to investigate the gradle call graph to find out what it's doing. Gradle is kind of magical.

PR #5, opened 2016-05-23
Test Cases for Scroll, Swipe, Longclick Handlers

anandsadu:
Test Cases for Scroll, Swipe, Longclick Handlers


jlipps:
from the little Java I know: 👍

PR #6, opened 2016-05-25
Code refactored and added Javadoc

anandsadu:
Code refactored and added Javadoc

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/handler/AppStrings.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/AppStrings.java
@@ -5,20 +5,20 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 
-import io.appium.uiautomator2.handler.request.BaseRequestHandler;
+import io.appium.uiautomator2.handler.request.SafeRequestHandler;
 import io.appium.uiautomator2.http.AppiumResponse;
 import io.appium.uiautomator2.http.IHttpRequest;
 import io.appium.uiautomator2.server.WDStatus;
 import io.appium.uiautomator2.utils.Logger;
 
-public class AppStrings extends BaseRequestHandler {
+public class AppStrings extends SafeRequestHandler {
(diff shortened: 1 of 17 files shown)


app/src/main/java/io/appium/uiautomator2/handler/AppStrings.java
```diff
@@ -5,20 +5,20 @@
...
     public AppStrings(String mappedUri) {
         super(mappedUri);
     }
 
     @Override
-    public AppiumResponse handle(IHttpRequest request) {
-        String jsonString = "";
+    public AppiumResponse safeHandle(IHttpRequest request) {
+        String jsonString;
         final String filePath = "/data/local/tmp/strings.json";
```
mpujari: Is there any other way to get filePath, as you can see its hardcoded.

app/src/main/java/io/appium/uiautomator2/handler/AppStrings.java
```diff
@@ -5,20 +5,20 @@
...
```
mpujari: Can we use try-with-resources for DataInputStream please.

anandsadu: try-with-resources is only supported if API level is set to 19 or higher. Currently our  min level is 18.

mpujari: Sorry, couldn't get you on that, what API level we are talking about here?

anandsadu: Java 7 feature ,try with resources is not allowing to compile the code. Since we have
 we have **minSdk** level 18 because our driver will support for **android API level 18** and above. If our minSdk is 19 then we would have used it.
https://github.com/appium/appium-uiautomator2-server/blob/master/app/build.gradle#L26

http://stackoverflow.com/questions/24290830/try-with-resources-are-not-supported-at-this-language-level-android

PR #10, opened 2016-06-14
Added Touch up,Touch move and Touch down handlers

anandsadu:
Added Touch up,Touch move and Touch down handlers.

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/handler/TouchEvent.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/TouchEvent.java
@@ -19,7 +19,6 @@ import io.appium.uiautomator2.utils.Logger;
 public abstract class TouchEvent extends SafeRequestHandler {
     protected int clickX, clickY;
     private AndroidElement element;
-    private boolean isElement;
 
     public TouchEvent(String mappedUri) {
         super(mappedUri);
@@ -67,7 +66,6 @@ public abstract class TouchEvent extends SafeRequestHandler {
         if (duration.length > 0) {
             extra = ", duration: " + duration[0];
         }
-        Logger.debug("Performing " + methodName + " using element? " + isElement
-                + " x: " + clickX + ", y: " + clickY + extra);
+        Logger.debug("Performing " + methodName + " x: " + clickX + ", y: " + clickY + extra);
     }
 }
--- a/app/src/main/java/io/appium/uiautomator2/handler/TouchEvent.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/TouchEvent.java
@@ -34,12 +34,12 @@ public abstract class TouchEvent extends SafeRequestHandler {
                 String id = json.getString("id");
                 element = KnownElements.getElementFromCache(id);
 
-                if (element != null)
-                    isElement = true;
-                if (isElement) {
+                if (element != null) {
                     final Rect bounds = element.getBounds();
                     clickX = bounds.centerX();
                     clickY = bounds.centerY();
+                } else {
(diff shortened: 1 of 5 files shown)


app/src/main/java/io/appium/uiautomator2/handler/TouchEvent.java
```diff
@@ -0,0 +1,73 @@
...
+        String path = "$.params.";
+        try {
+            JSONObject json = getPayload(request);
+            if (json.has("id")) {
+                String id = json.getString("id");
+                element = KnownElements.getElementFromCache(id);
+
+                if (element != null)
+                    isElement = true;
+                if (isElement) {
```
imurchie: Is it possible to get here and not have an element? With the logic as it stands, if we get to here and there is no element, we will not populate `clickX` and `clickY` at all. Not sure what the consequences of that would be, though it seems like the touch event would not work.

anandsadu: Yeah..Here  I have added the exception for the missed case. If we have element we will get the bounds else we can get the x and y  coordinates from the json.
```
          if (json.has("id")) {
                String id = json.getString("id");
                element = KnownElements.getElementFromCache(id);
                if (element != null) { 
                       final Rect bounds = element.getBounds();
                       clickX = bounds.centerX();
                       clickY = bounds.centerY();
                 } else {
                    throw new UiObjectNotFoundException("Touch Event Element doesn't exist");
                    }
                }               
            } else { // no element so extract x and y from params
                clickX = JsonPath.compile(path + "x").read(json.toString());
                clickY = JsonPath.compile(path + "y").read(json.toString());
            }
```

imurchie: Where is this code from?

anandsadu: Referred [android bootstrap ](https://github.com/appium/appium-android-bootstrap/blob/master/bootstrap/src/io/appium/android/bootstrap/handler/TouchEvent.java#L57). I'll test the code with the changes mentioned in above comment and commit.

app/src/main/java/io/appium/uiautomator2/handler/TouchEvent.java
```diff
@@ -0,0 +1,73 @@
...
+            if (json.has("id")) {
+                String id = json.getString("id");
+                element = KnownElements.getElementFromCache(id);
+
+                if (element != null) {
+                    final Rect bounds = element.getBounds();
+                    clickX = bounds.centerX();
+                    clickY = bounds.centerY();
+                } else {
+                    throw new UiObjectNotFoundException("TouchEvent element does not exist.");
```
imurchie: Do you want to log the element id that was attempted?

app/src/main/java/io/appium/uiautomator2/handler/TouchEvent.java
```diff
@@ -0,0 +1,73 @@
...
+
+    protected abstract boolean executeTouchEvent() throws UiObjectNotFoundException;
+
+    protected void printEventDebugLine(final String methodName,
+                                       final Integer... duration) {
+        String extra = "";
+        if (duration.length > 0) {
+            extra = ", duration: " + duration[0];
+        }
+        Logger.debug("Performing " + methodName + " using element? " + isElement
```
imurchie: `isElement` is no longer being set.

anandsadu: yeah...will remove that

PR #12, opened 2016-06-21
Modified Swipe handler and integrated with driver module

anandsadu:
Modified Swipe handler  according to the json from the driver. Driver changes [here](https://github.com/appium/appium-uiautomator2-driver/pull/25)


imurchie:
LGTM

PR #13, opened 2016-06-21
Code refactor

sravanmedarapu:
- Updated throwing RuntimeException to more appropriate exceptions.
Reason for change: Netty server explicitly catching all RuntimeExceptions and throwing it's own custom messages like **server hangup** and external modules who calls server don't have control actual exception.

- Updated UiAutomator2Exception to extend Exception from RuntimeException

- Updated [CustomUIDevice](https://github.com/sravanmedarapu/appium-uiautomator2-server/blob/code_refactor/app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java#L178).
Reason for updated: `mInstrumentation.getUiAutomation().getRootInActiveWindow()` returning `null` randomly, need to figure out actual root cause, for time being repolling for maximum 5 times for getting root node for being non null.

### DIFF
--- a/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/TestHelper.java
+++ b/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/TestHelper.java
@@ -18,7 +18,6 @@ import org.json.JSONException;
 
 import java.io.IOException;
 
-import io.appium.uiautomator2.common.exceptions.UiAutomator2Exception;
 import io.appium.uiautomator2.utils.Logger;
 
 import static android.os.SystemClock.elapsedRealtime;
@@ -74,7 +73,7 @@ public abstract class TestHelper {
             Response response = client.newCall(request).execute();
             result = response.body().string();
         } catch (IOException e) {
-            throw new UiAutomator2Exception(request.method() + " \"" + request.urlString() + "\" failed. " + e);
+            throw new RuntimeException(request.method() + " \"" + request.urlString() + "\" failed. " + e);
         }
         return result;
     }
--- a/app/src/main/java/io/appium/uiautomator2/core/UiAutomatorBridge.java
+++ b/app/src/main/java/io/appium/uiautomator2/core/UiAutomatorBridge.java
@@ -48,10 +48,10 @@ public class UiAutomatorBridge {
 
             this.uiAutomatorBridge = getField(UiDevice.class, FIELD_UI_AUTOMATOR_BRIDGE, device);
         } catch (Error error) {
-            Logger.error("ERROR", "error", error);
+            Logger.error("ERROR", error);
             throw error;
         } catch (UiAutomator2Exception error) {
-            Logger.error("ERROR", "error", error);
+            Logger.error("ERROR", error);
             throw new Error(error);
         }
     }
--- a/app/src/main/java/io/appium/uiautomator2/core/UiAutomatorBridge.java
+++ b/app/src/main/java/io/appium/uiautomator2/core/UiAutomatorBridge.java
@@ -20,6 +20,7 @@ import android.util.Log;
 import android.view.Display;
 import android.view.InputEvent;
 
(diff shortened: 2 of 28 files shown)


bootstraponline:
👍

app/src/main/java/io/appium/uiautomator2/core/UiAutomatorBridge.java
```diff
@@ -49,26 +50,29 @@ public UiAutomatorBridge() {
         } catch (Error error) {
             Log.e("ERROR", "error", error);
             throw error;
+        } catch (UiAutomator2Exception error) {
+            Log.e("ERROR", "error", error);
```
imurchie: Elsewhere in this PR you've changed `Log.e` to `Log.error`. Not sure of the reason for that, but does it apply here too?

app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
```diff
@@ -166,7 +166,33 @@ public Object findObject(Object selector) throws ClassNotFoundException, Element
             }
             // Prior to API level 21 we can only access the active window
         } else {
-            ret.add(mInstrumentation.getUiAutomation().getRootInActiveWindow());
+            AccessibilityNodeInfo node = mInstrumentation.getUiAutomation().getRootInActiveWindow();
+            if (node != null) {
+                ret.add(node);
+            } else {
+                /*TODO: As we can't proceed to find element with out root node,
```
imurchie: Why are these TODOs? They read like comments on the following code.

sravanmedarapu: Current implementation is a workaround, need to investigate more for the route cause of the issue and for better understanding mentioned wrote little in detail.

app/src/main/java/io/appium/uiautomator2/core/UiAutomatorBridge.java
sravanmedarapu: Sure, will update to Logger.error :+1:

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/TestHelper.java
```diff
@@ -74,7 +73,7 @@ private static String execute(Request request) {
             Response response = client.newCall(request).execute();
             result = response.body().string();
         } catch (IOException e) {
-            throw new UiAutomator2Exception(request.method() + " \"" + request.urlString() + "\" failed. " + e);
+            throw new RuntimeException(request.method() + " \"" + request.urlString() + "\" failed. " + e);
```
mpujari: Do we want to use ```throw new RuntimeException(<string>, e) instead.

app/src/main/java/io/appium/uiautomator2/handler/Source.java
```diff
@@ -30,24 +31,25 @@ public Source(String mappedUri) {
...
+            Transformer transformer = tf.newTransformer();
             transformer.transform(new DOMSource(doc), new StreamResult(writer));
-            xmlString = writer.getBuffer().toString();
+            String xmlString = writer.getBuffer().toString();
             return new AppiumResponse(getSessionId(request), WDStatus.SUCCESS, xmlString);
+
         } catch (final TransformerConfigurationException e) {
             Logger.error("Unable to handle the request:" + e);
-            throw new RuntimeException("Something went terribly wrong while converting xml document to string", e);
+            return new AppiumResponse(getSessionId(request), WDStatus.UNKNOWN_ERROR, "Something went terribly wrong while converting xml document to string:" + e);
```
mpujari: Same comment from above, do we want to use ```, e)``` instead of ```+ e```

```
return new AppiumResponse(getSessionId(request), WDStatus.UNKNOWN_ERROR, "Something went terribly wrong while converting xml document to string:", e);
```

PR #14, opened 2016-06-29
Code refactor

sravanmedarapu:
-  Updated json body for FindElement/FindElements and element based actions like (click, clear, flick, e.t.c) to be in sync with driver module

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/http/AppiumResponse.java
+++ b/app/src/main/java/io/appium/uiautomator2/http/AppiumResponse.java
@@ -39,6 +39,7 @@ public class AppiumResponse {
         } catch (JSONException e) {
             Logger.error("Unable to create JSON Object:", e);
         }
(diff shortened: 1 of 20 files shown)


app/src/main/java/io/appium/uiautomator2/http/AppiumResponse.java
```diff
@@ -39,6 +39,7 @@ public String render() {
         } catch (JSONException e) {
             Logger.error("Unable to create JSON Object:", e);
         }
+        Logger.info("AppiumResponse: ", o.toString());
```
mpujari: Do we want to use debug level here rather then info.

sravanmedarapu: Used `info` as it is a general information for end user.

mpujari: Not sure but it seems its more dev info rather then user.

PR #16, opened 2016-07-15
500 HTTP status for unsuccessful requests

sravanmedarapu:
- For all non '0' status codes returning 500 HTTP status

- Added corresponding test


imurchie:
This looks good to me.

PR #18, opened 2016-07-19
Handled Exceptions and refactored Code

anandsadu:


app/src/main/java/io/appium/uiautomator2/handler/Clear.java
```diff
@@ -25,7 +25,11 @@ public AppiumResponse safeHandle(IHttpRequest request) {
             JSONObject payload = getPayload(request);
             String id = payload.getString("elementId");
             AndroidElement element = KnownElements.getElementFromCache(id);
-            element.clear();
+            if (element != null) {
+                element.clear();
+            } else {
```
imurchie: In all of these it seems cleaner to me to reverse the check for `null`, so the error is handled and then the actual work is done outside of the conditional.

```java
if (element == null) {
    return new AppiumResponse(getSessionId(request), WDStatus.NO_SUCH_ELEMENT, "Element Not found");
}

element.clear();
```

This doesn't affect much here, but in other cases below where the action is longer, it makes it clearer what the error condition is. It also gets rid of lots of indenting. :)

imurchie:
👍

PR #20, opened 2016-08-01
Refactored Swipe json body and corresponding test

sravanmedarapu:


app/src/main/java/io/appium/uiautomator2/handler/Swipe.java
```diff
@@ -20,19 +19,18 @@ public Swipe(String mappedUri) {
     @Override
     public AppiumResponse safeHandle(IHttpRequest request) {
         try {
-            String json = getPayload(request).toString();
+            JSONObject payload = getPayload(request);
```
imurchie: 👍

PR #21, opened 2016-08-04
Added Swipe on the Element functionality along with coordinates

anandsadu:


app/src/main/java/io/appium/uiautomator2/handler/Swipe.java
```diff
@@ -1,51 +1,91 @@
...
 public class Swipe extends SafeRequestHandler {
+    UiDevice device;
 
     public Swipe(String mappedUri) {
         super(mappedUri);
     }
 
     @Override
     public AppiumResponse safeHandle(IHttpRequest request) {
+        Point absStartPos, absEndPos;
```
mpujari: Minor: declare the variable where its being first initialized, by this we narrow the scope of the variable. Same goes with swipeArgs as well.

app/src/main/java/io/appium/uiautomator2/handler/Swipe.java
```diff
@@ -1,51 +1,91 @@
...
-            // TODO Swipe on Element
-
-            int startX = payload.getInt("startX");
-            int startY = payload.getInt("startY");
-            int endX = payload.getInt("endX");
-            int endY = payload.getInt("endY");
-            int steps = payload.getInt("steps");
+            Logger.info("JSON Payload : ", payload.toString());
+            swipeArgs = new SwipeArguments(request);
+            device = Device.getUiDevice();
```
mpujari: Any reason we kept **deivce** as instance variable? but looking at the code it seems we don't require it.

anandsadu: Yeah.Not required as instance variable .Will make the change.Thanks

PR #22, opened 2016-08-08
Code refactor

sravanmedarapu:
- Added AppiumResponse constructor to accept throwable type

- Updated FindElement & FindElements's By.Id to add package name as prefix for not a fully qualified id

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/handler/FindElement.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/FindElement.java
@@ -10,6 +10,7 @@ import org.json.JSONObject;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
+import java.util.regex.Pattern;
 
 import javax.xml.parsers.ParserConfigurationException;
 
@@ -25,6 +26,7 @@ import io.appium.uiautomator2.model.By;
 import io.appium.uiautomator2.model.By.ByClass;
 import io.appium.uiautomator2.model.By.ById;
 import io.appium.uiautomator2.model.KnownElements;
+import io.appium.uiautomator2.model.Session;
 import io.appium.uiautomator2.model.internal.NativeAndroidBySelector;
 import io.appium.uiautomator2.server.WDStatus;
 import io.appium.uiautomator2.utils.ClassInstancePair;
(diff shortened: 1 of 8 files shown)


app/src/main/java/io/appium/uiautomator2/handler/FindElement.java
```diff
@@ -38,6 +40,21 @@
...
+     * com.example.Test:id/enter
+     *
+     * ^[a-zA-Z_] - Java package must start with letter or underscore
+     * [a-zA-Z0-9\._]* - Java package may contain letters, numbers, periods and
+     * underscores : - : ends the package and starts the type [^\/]+ - type is
+     * made up of at least one non-/ characters \\/ - / ends the type and starts
+     * the name [\S]+$ - the name contains at least one non-space character and
+     * then the line is ended
+     */
+    static final Pattern resourceIdRegex   = Pattern
```
imurchie: I find it very helpful to add, in a comment, a link to a regular expression testing site with the regex and expected data. Not sure of the best one for Java

app/src/main/java/io/appium/uiautomator2/handler/FindElements.java
```diff
@@ -121,6 +122,16 @@ public AppiumResponse safeHandle(IHttpRequest request) {
 
     private List<Object> findElements(By by) throws ElementNotFoundException, ParserConfigurationException, ClassNotFoundException, InvalidSelectorException, UiAutomator2Exception {
         if (by instanceof By.ById) {
+            String locator = by.getElementLocator();
```
imurchie: This whole section is repeated from `FindElement`. Is there no way to use the version from there without copying it?

app/src/main/java/io/appium/uiautomator2/handler/FindElement.java
sravanmedarapu: Sure, will add.

app/src/main/java/io/appium/uiautomator2/handler/NewSession.java
```diff
@@ -30,4 +37,19 @@ public AppiumResponse safeHandle(IHttpRequest request) {
         return new AppiumResponse(sessionID, WDStatus.SUCCESS, "Created Session");
     }
 
+
+    public  Map<String, Object> getCapabilities(IHttpRequest request) throws JSONException {
+        JSONObject caps = getPayload(request).getJSONObject("desiredCapabilities");
+        Map<String, Object> map = new HashMap<String, Object>();
+
+        Iterator<String> keysItr = caps.keys();
+        while(keysItr.hasNext()) {
```
mpujari: Minor: Can we use below pattern

```
for (Map.Entry<String, Object> entry : caps.entrySet()) {
   map.put(entry.getKey(), entry.getValue());
}
```

app/src/main/java/io/appium/uiautomator2/handler/NewSession.java
```diff
@@ -30,4 +37,19 @@ public AppiumResponse safeHandle(IHttpRequest request) {
         return new AppiumResponse(sessionID, WDStatus.SUCCESS, "Created Session");
     }
 
+
+    public  Map<String, Object> getCapabilities(IHttpRequest request) throws JSONException {
+        JSONObject caps = getPayload(request).getJSONObject("desiredCapabilities");
+        Map<String, Object> map = new HashMap<String, Object>();
+
```
mpujari: Instead of using iterator, get and next, can we use below code

```
for (Map.Entry<String, Object> entry : caps.entrySet()) {
   map.put(entry.getKey(), entry.getValue());
}
```

sravanmedarapu: seems `entrySet() on JSONObject` available in `GSON(com.google.gson.JsonObject)` library but here we are using `org.json.JSONObject`

PR #24, opened 2016-08-24
Added tap end point and updated .npmignore

sravanmedarapu:
- Added Tap Endpoint

- Added Test for TouchEvents

- Refactored Tests & DeleteSession handler

- Updated .npmignore to ignore local.properties file

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
+++ b/app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
@@ -3,8 +3,12 @@ package io.appium.uiautomator2.server;
 import android.content.Context;
 import android.os.Looper;
 import android.os.PowerManager;
+import android.os.RemoteException;
 
+import io.appium.uiautomator2.common.exceptions.SessionRemovedException;
 import io.appium.uiautomator2.utils.Logger;
+
+import static io.appium.uiautomator2.utils.Device.getUiDevice;
 
 public class ServerInstrumentation {
     private static ServerInstrumentation instance;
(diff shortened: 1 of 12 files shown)


app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
```diff
@@ -3,15 +3,20 @@
...
 
+import static io.appium.uiautomator2.utils.Device.getUiDevice;
+
 public class ServerInstrumentation {
     private static ServerInstrumentation instance;
     private static Context context;
     private HttpdThread serverThread;
     private PowerManager.WakeLock wakeLock;
     private int serverPort = 8080;
+    private static boolean isStopServer;
```
imurchie: What does this track? It seems to only be set once, to `True`, and then can never be anything else.

imurchie:
One question. Otherwise nothing jumps out at me as odd.

app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
sravanmedarapu: We are tracking the server status[`isStopServer()`] in [Test](https://github.com/sravanmedarapu/appium-uiautomator2-server/blob/7228aef39b66ec357d1aae065f0283ca80284c5b/app/src/androidTestServer/java/io/appium/uiautomator2/server/test/AppiumUiAutomator2Server.java#L34), once `isStopServer` status is`True` we are completing the Test.

imurchie: I see that. But since this is static and all, is it never dealt with anywhere else? It just seems weird to my eyes.

sravanmedarapu: Removed static property.

PR #26, opened 2016-09-30
Changed netty server port number and gradle dependency versions

sravanmedarapu:


imurchie:
Seems reasonable to me!

PR #31, opened 2016-10-19
Disabled mutli-window and added test

sravanmedarapu:
- Disabled finding element on multi window based AccessibilityWindowInfo


imurchie:
This seems fine to me!

PR #37, opened 2016-11-16
Added TOAST message verification support

sravanmedarapu:
- Added Toast message verification support. #https://github.com/appium/appium/issues/4824

    1.  Usage example with xpath `//*[@text='Clicked popup menu item Add']`

    1.  It is only possible to find the element, we can't perform any operation on TOAST element. Performing any actions on TOAST element will always throw `StaleObjectException`.

- Refactored `LongTouch` json body.

- Code cleanup - avoid creating new `UiDevice` instances.


samsonbarukula:
can anyone provide toast message java sample code please thank you

PR #47, opened 2017-01-02
touch/flick payload compliance

sa-spag:
According to [the initial specification](https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidtouchflick), `POST session/:sessionId/touch/flick` expects `element` as a JSON parameter instead of `elementId`.


sa-spag:
I believe it has been fixed with https://github.com/appium/appium-uiautomator2-server/commit/c4b48a1e698955d3c5582edec4c9612ca08460ff.

sravanmedarapu:
@sa-spag, Yeah this has been fixed and should be available in appium-uiautomator2-driver v0.2.7.

However thanks for the PR, I really appreciate your efforts :)

PR #48, opened 2017-01-03
command handler to look for "element" instead of "elementId"

meprabhatr:
[Problem]
Appium elementId in command with keyname "element". UIAutomator server commands
expect keyname "elementId" which causes failure in command execution.
related: https://github.com/appium/appium-base-driver/pull/96

[Solution]
Using keyname "element" instead of "elementId". Updated code and test

[Test]
Verified that tests work fine.

removing wrong commit

### DIFF
--- a/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/HandlersTest.java
+++ b/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/HandlersTest.java
@@ -180,7 +180,7 @@ public class HandlersTest {
         String destElementId = new JSONObject(new JSONObject(destElement).get("value").toString()).get("ELEMENT").toString();
 
         JSONObject dragBody = new JSONObject();
-        dragBody.put("elementId", srcElementId);
+        dragBody.put("element", srcElementId);
         dragBody.put("destElId", destElementId);
         dragBody.put("startX", startX);
         dragBody.put("startY", startY);
--- a/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/TestUtil.java
+++ b/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/TestUtil.java
@@ -155,7 +155,7 @@ public class TestUtil {
         }
 
         JSONObject jsonObject = new JSONObject();
-        jsonObject.put("elementId", elementId);
+        jsonObject.put("element", elementId);
         return post(baseUrl + "/element/" + elementId + "/click", jsonObject.toString());
     }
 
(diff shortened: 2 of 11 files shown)


sravanmedarapu:
`elementId` property is tightly coupled with appium-uiautomator2-driver module [element.js](https://github.com/appium/appium-uiautomator2-driver/blob/bc61cdee16ff92494947760e8385c598e83d1e2d/lib/commands/element.js) and [touch.js](https://github.com/appium/appium-uiautomator2-driver/blob/bc61cdee16ff92494947760e8385c598e83d1e2d/lib/commands/touch.js), we should also make necessary changes in [driver](https://github.com/appium/appium-uiautomator2-driver) to be in sync with server.

meprabhatr:
Hi @sravanmedarapu, I did not quite get what changes in driver code you are referring to? elementId in element.js and touch.js is a variable which holds the value of element id and is appended to command URI.
The change in this CR is for case when http requests are proxied to server with payload([proxyReqRes](https://github.com/appium/appium-base-driver/blob/master/lib/jsonwp-proxy/proxy.js#L163) in [appium-base-driver.](https://github.com/appium/appium-base-driver/))
Along with other key-value pair, payload has  key-value "element":elementId but server was expecting key-value "elementId":elementId. So the change in server is made to look for key-value "element":elementId in payload.

meprabhatr:
Got it, thanks! 
Will just replacing all "elementId" with "element" in files element.js and touch.js be sufficient?

sravanmedarapu:
@meprabhatr, we need to update [`touchLongClick`](https://github.com/appium/appium-uiautomator2-driver/blob/master/lib/commands/element.js#L31), [`touchDown`](https://github.com/appium/appium-uiautomator2-driver/blob/master/lib/commands/element.js#L36), [`touchUp`](https://github.com/appium/appium-uiautomator2-driver/blob/master/lib/commands/element.js#L41), [`touchMove`](https://github.com/appium/appium-uiautomator2-driver/blob/master/lib/commands/element.js#L46), [`click`](https://github.com/appium/appium-uiautomator2-driver/blob/master/lib/commands/element.js#L59) and [`tap`](https://github.com/appium/appium-uiautomator2-driver/blob/master/lib/commands/element.js#L67).

we also need to revert changes in [Swipe](https://github.com/appium/appium-uiautomator2-server/pull/48/files#diff-9b8f80580f543fc8e3a8ea63764d7bd7R37), [SendKeysToElement](https://github.com/appium/appium-uiautomator2-server/pull/48/files#diff-ee9cc3194d75ef0c24a77fa4d1aef97dR36) and its corresponding tests as JSON body for these handlers depends on android-driver [setElementValue](https://github.com/appium/appium-android-driver/blob/d615b15319f00181a053664f5ed6f68338a9fd2b/lib/commands/element.js#L39) and [swipe](https://github.com/appium/appium-android-driver/blob/25b3d5271d1aa3b38cdddb9c3f07ce40a7840b49/lib/commands/actions.js#L62) on which we don't have control.

meprabhatr:
@sravanmedarapu, changes made as suggested.

sravanmedarapu:
@meprabhatr, Thanks for the PR. I appreciate your efforts :)

PR #56, opened 2017-05-05
The returned value for boolean argument types should be of type string

mykola-mokhnach:
Fixes https://github.com/appium/appium/issues/8321

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
@@ -38,7 +38,10 @@ public class GetElementAttribute extends SafeRequestHandler {
                 return new AppiumResponse(getSessionId(request), WDStatus.SUCCESS, attribute);
             } else {
                 Boolean boolAttribute = element.getBoolAttribute(attributeName);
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
```diff
@@ -38,7 +38,10 @@ public AppiumResponse safeHandle(IHttpRequest request) {
                 return new AppiumResponse(getSessionId(request), WDStatus.SUCCESS, attribute);
             } else {
                 Boolean boolAttribute = element.getBoolAttribute(attributeName);
-                return new AppiumResponse(getSessionId(request), WDStatus.SUCCESS, boolAttribute);
+                // The result should be anyway of type string according to
```
jlipps: NIT: no need for the word `anyway` here?

mykola-mokhnach: natürlich

sravanmedarapu:
Would be great if we also update the corresponding [tests](https://github.com/appium/appium-uiautomator2-server/blob/master/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/HandlersTest.java#L773).

mykola-mokhnach:
@sravanmedarapu done

Issue #61, opened 2017-06-07
Unable to find android system resources like statusBarBackground


rjaros87:
Hi,
After upgrading Appium from 1.6.3 (uiautomator2-server v0.0.8) to Appium 1.6.5 (uiautomator2-server v0.1.5) I'm unable to find android bars like `android:id/statusBarBackground` or `android:id/navigationBarBackground` during my test which requires to interact with this bars.

The commit https://github.com/appium/appium-uiautomator2-server/commit/5a7adfb5669cf770a73c34240ec7c6b64868a944 cause this behavior.

Should this be fixed? or exposed some settings in appium capabilities to control `isVisibleToUser`.

The statusBar and navigationBar is displayed and I don't know why android `isVisibleToUser` decided is not visible.

PR #63, opened 2017-06-13
Allow invisible elements

rjaros87:
Solves issue #61


rjaros87:
@sravanmedarapu 

> driver.updateSettings({'allowInvisibleElements': 'true'})

If everything is ok, I will prepare PR for appium-uiautomator2-driver.

app/src/main/java/io/appium/uiautomator2/server/ServerConfig.java
```diff
@@ -4,6 +4,25 @@
 public class ServerConfig {
     private final static int PORT = 6790;
 
+    private static boolean allowInvisibleElements = false;
```
sravanmedarapu: I think adding them to [session](https://github.com/appium/appium-uiautomator2-server/blob/8774414fdd6b5b3b846039af12de78c9bd3b2bbd/app/src/main/java/io/appium/uiautomator2/model/Session.java#L15) would be a better approach, because down the line if there is any new setting to add, then we need to add that attribute and also need to corresponding setter and getter methods. Instead, if we can add them to session then it will minimize the effort.

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
```diff
@@ -103,6 +104,7 @@ private void registerPostHandler() {
         register(postHandler, new TouchDown("/wd/hub/session/:sessionId/touch/down"));
         register(postHandler, new TouchUp("/wd/hub/session/:sessionId/touch/up"));
         register(postHandler, new TouchMove("/wd/hub/session/:sessionId/touch/move"));
+        register(postHandler, new AllowInvisibleElements("/wd/hub/session/:sessionId/appium/device/allowInvisibleElements"));
```
sravanmedarapu: Please updating endpoint to `/wd/hub/session/:sessionId/appium/settings` same way as the appium [API](https://github.com/appium/appium-base-driver/blob/master/lib/mjsonwp/routes.js#L430-L432), rename `AllowInvisibleElements` to `UpdateSettings` and add property to [session](https://github.com/appium/appium-uiautomator2-server/blob/8774414fdd6b5b3b846039af12de78c9bd3b2bbd/app/src/main/java/io/appium/uiautomator2/model/Session.java#L15)

sravanmedarapu:
@rjaros87, thanks for the quick PR. Looks good, couple of minor suggestion :)

rjaros87:
Please review :)
In `appium-uiautomator2-driver`  i will update `onSettingsUpdate` like below
```
async onSettingsUpdate (key, value) {
    let newSettings = {};
    newSettings[key] = value;
    await this.uiautomator2.jwproxy.command('/appium/settings', 'POST', {settings: newSettings});
  }
```
is this what you mean?

jlipps:
@rjaros87 fyi you can tighten up the JS code a bit that you will write:

```js
async onSettingsUpdate (key, value) {
    let settings = {[key]: value};
    await this.uiautomator2.jwproxy.command('/appium/settings', 'POST', {settings});
}
```

PR #65, opened 2017-07-09
Allow null values in JSON response

mykola-mokhnach:
JSONWP proxy contains the following verification for JSON response:

```js
    resBody = util.safeJsonParse(resBody);
    if (_.includes(statusCodesWithRes, response.statusCode) &&
        (_.isUndefined(resBody.status) || _.isUndefined(resBody.value))) {
      throw new Error(`Did not get a valid response object. Object was: ${JSON.stringify(resBody)}`);
    }
```

This means we get an exception even in case of correct response, since the valid _null_ value is deliberately not included there inside server code. Now this is fixed.

Addresses https://github.com/appium/appium/issues/8810

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/http/AppiumResponse.java
+++ b/app/src/main/java/io/appium/uiautomator2/http/AppiumResponse.java
@@ -45,9 +45,7 @@ public class AppiumResponse {
                 o.put("sessionId", sessionId);
             }
             o.put("status", status);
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
@sravanmedarapu Any change to get this fix merged?

sravanmedarapu:
@mykola-mokhnach, sorry for the delayed response. look good 👍

PR #80, opened 2017-10-12
set waitForIdleTimeout via settings

ajeeshvl:
This is to set the waitForIdleTimeout via settings, which will help user to override the default Idle timeout(10 *1000).
For example, If there are elements which only appears in the screen for less than 5 seconds. In a video player, player controls overlay disappears in 5 seconds. UIAutomator2 waits for screen to become idle and failed to identify the element since there is a default idle timeout of 10 seconds.
This resolves the issue appium/appium#9147

From clients, We can set settings like driver.setSetting(Setting.WAIT_FOR_IDLE_TIMEOUT, 0);

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/model/WaitForIdleTimeout.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/WaitForIdleTimeout.java
@@ -13,7 +13,7 @@ public class WaitForIdleTimeout {
       Configurator.getInstance().setWaitForIdleTimeout(timeout);
       Logger.debug("Set waitForIdleTimeout to: " + timeout);
     } catch (Exception e) {
-      Logger.error("error setting waitForIdleTimeout " + e.getMessage());
+      Logger.error("Error setting waitForIdleTimeout " + e.getMessage());
     }
   }
 }
--- a/app/src/main/java/io/appium/uiautomator2/model/WaitForIdleTimeout.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/WaitForIdleTimeout.java
@@ -1,0 +1,19 @@
+package io.appium.uiautomator2.model;
+
+import android.support.test.uiautomator.Configurator;
+
+import io.appium.uiautomator2.utils.Logger;
+
+public class WaitForIdleTimeout {
+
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/appium/uiautomator2/model/WaitForIdleTimeout.java
```diff
@@ -0,0 +1,19 @@
+package io.appium.uiautomator2.model;
+
+import android.support.test.uiautomator.Configurator;
+
+import io.appium.uiautomator2.utils.Logger;
+
+public class WaitForIdleTimeout {
+
+  public static final String SETTING_NAME = "setWaitForIdleTimeout";
```
vmaxim: I would suggest to remove `set` prefix from setting name.

ajeeshvl: Client implementation uses enum constant as "setWaitForIdleTimeout" 
Java_Client:
https://github.com/appium/java-client/blob/37f5ed3677edc68370e98911c7ef89a3c8ee9c1c/src/main/java/io/appium/java_client/Setting.java

appium-dotnet-driver:
https://github.com/appium/appium-dotnet-driver/blob/698351d7e301dc1acb09a5b333907d96686c4281/appium-dotnet-driver/Appium/Android/Enums/AutomatorSetting.cs

vmaxim: I don't think it works on client side, because only `ignoreUnimportantViews` implemented in [appium-android-driver](https://github.com/appium/appium-android-driver/blob/master/lib/driver.js#L46-L47) and plus `allowInvisibleElements` in [appium-uiautomator2-driver](https://github.com/appium/appium-uiautomator2-driver/blob/c2133948a084a6d122e1f1448211e9754bcdc3c5/lib/driver.js#L111-L112).
Also you can take a look at additional settings implemented in [appium-android-bootstrap](https://github.com/appium/appium-android-bootstrap/blob/e0df190e35baccf31b4e2af81db75cff580631a1/bootstrap/src/io/appium/android/bootstrap/handler/ConfiguratorHandler.java#L21-L25) but not implemented in driver.

PR #89, opened 2017-11-09
Replace `clearTextField()` with `setText('')`

vmaxim:
`UiObject.clearTextField()` method [can not clear password text field](https://android.googlesource.com/platform/frameworks/uiautomator/+/android-support-test/src/main/java/android/support/test/uiautomator/UiObject.java#693) because `getText()` returns empty string, but we can mimic [`UiObject2.clear()`](https://android.googlesource.com/platform/frameworks/uiautomator/+/android-support-test/src/main/java/android/support/test/uiautomator/UiObject2.java#337) behaviour.
Related to appium/appium-uiautomator2-driver#109

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/model/UiObjectElement.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/UiObjectElement.java
@@ -128,7 +128,7 @@ public class UiObjectElement implements AndroidElement {
     }
 
     public void clear() throws UiObjectNotFoundException {
-        element.clearTextField();
+        element.setText("");
     }
 
     public String getId() {


app/src/main/java/io/appium/uiautomator2/model/UiObjectElement.java
```diff
@@ -128,7 +128,7 @@ public By getBy() {
     }
 
     public void clear() throws UiObjectNotFoundException {
-        element.clearTextField();
+        element.setText("");
```
mykola-mokhnach: shall we check if this is really a password field first?

vmaxim: `UiObject2.clear()` always perform `setText("")`. I think it can be a bit confusing if the behavior depends on the type of element and locator.

mykola-mokhnach: I'm just wondering if it does not generate some additional events or something like that as a part of the method call. If you think that's safe to replace the current call and you tested it then no problems.

vmaxim: There is no any difference for API>19, except [checking element text length](https://android.googlesource.com/platform/frameworks/uiautomator/+/android-support-test/src/main/java/android/support/test/uiautomator/UiObject.java#693)

PR #90, opened 2017-11-10
Use "live" AccessibilityNodeInfo instead of cached

vmaxim:


app/src/main/java/io/appium/uiautomator2/model/UiObject2Element.java
```diff
@@ -60,6 +60,7 @@ public String getText() throws UiObjectNotFoundException {
         if (isToastElement(nodeInfo)) {
             return nodeInfo.getText().toString();
         }
+        nodeInfo = AccessibilityNodeInfoGetter.fromUiObject(element);
```
mykola-mokhnach: what is the point of doing this?

vmaxim: Because [getAccessibilityNodeInfo()](https://android.googlesource.com/platform/frameworks/uiautomator/+/android-support-test/src/main/java/android/support/test/uiautomator/UiObject2.java#619) refresh element's accessiblity info.
Without refreshing it returns old cached value of [getRangeInfo()](https://github.com/appium/appium-uiautomator2-server/blob/master/app/src/main/java/io/appium/uiautomator2/model/UiObject2Element.java#L64)

mykola-mokhnach: then it makes sense to do this inside `ff` block

mykola-mokhnach: laso add a comment why it is needed

PR #105, opened 2018-01-08
Extend the range of unacceptable characters for tag names 

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/9934


mykola-mokhnach:
@imurchie The log message is written now if `name != fixedName`

PR #107, opened 2018-01-10
Streamline publishing

imurchie:
Publishing now becomes a matter of running the normal flow. `npm version major|minor|patch` will build the apks and put them in a place known to the index. `npm publish` will create a tarball that includes the corresponding apks for use.

Usage would become a matter of importing the package and getting the `SERVER_APK_PATH` and `TEST_APK_PATH`. E.g., in the `appium-uiautomator2-driver` package,
```js
// -import { UI2_SERVER_APK_PATH as apkPath, UI2_TEST_APK_PATH as testApkPath, UI2_VER} from './installer';
import { UI2_VER } from './installer';
import { SERVER_APK_PATH as apkPath, TEST_APK_PATH as testApkPath } from 'appium-uiautomator2-server';
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "0.4.1",
   "description": "A netty server with uiautomator2 handlers",
   "main": "index.js",
-  "windowsBuildDir":"c:/tmp",
-   "repository": {
+  "windowsBuildDir": "c:/tmp",
+  "repository": {
     "type": "git",
     "url": "https://github.com/appium/appium-uiautomator2-server"
   },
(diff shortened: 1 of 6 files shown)


update-gradle-version.js
```diff
@@ -0,0 +1,36 @@
...
+  console.error('No updated version specified');
+  process.exit(1);
+}
+if (!/\d+\.\d+\.\d+/.test(version)) {
+  console.error(`Invalid version specified '${version}'`);
+  console.error(`Version should be in the form '1.2.3'`);
+  process.exit(2);
+}
+
+console.log(`Updating gradle build file to version '${version}'`);
```
mykola-mokhnach: I would also increase versionCode

update-gradle-version.js
```diff
@@ -0,0 +1,36 @@
...
+replace({
+  files: GRADLE_BUILD_FILE,
+  from: /^\s+versionName\s"(.+)"$/gm,
+  to: (match, ...args) => {
+    // match will be like `versionName "1.2.3"`
+    return match.replace(/\d+\.\d+\.\d/, version);
```
mykola-mokhnach: should we check that the new version is greater than the current one?

update-gradle-version.js
imurchie: That's not currently part of the publishing process.

update-gradle-version.js
imurchie: No. The version is provided by npm. It can only go up.

package.json
```diff
@@ -18,5 +18,17 @@
   "bugs": {
     "url": "https://github.com/appium/appium-uiautomator2-server/issues"
   },
-  "homepage": "https://github.com/appium/appium-uiautomator2-server"
+  "homepage": "https://github.com/appium/appium-uiautomator2-server",
+  "scripts": {
+    "bump-gradle-version": "node update-gradle-version.js --package-version=${npm_package_version} && git add app/build.gradle",
```
jlipps: awesome to have this be part of `npm version`!

PR #108, opened 2018-01-12
Shouln't use hardcoded timeout for waitForIdle().

SaneBow:
The `Device.waitForIdle` in `XPathFinder.java` shouldn't use hardcoded value 10*1000 ms. It should be UiAutomator who decide whether use default timeout value or read from user's setting. 

This also related to https://github.com/appium/appium/issues/9147. When there are some continues background animation, user may try to set a small `waitForIdleTimeout` to prevent it from hanging. But if he try to search XPath, it will seem like the setting has no effect.


imurchie:
Thanks for this!

imurchie:
Published in `appium-uiautomator2-server@1.0.3`.

PR #112, opened 2018-01-19
Speed up screenshoting

mykola-mokhnach:
Actually, the previous screenshoting endpoint was not working properly at all, since it was not returning Base64-encoded screenshot string. Now it's fixed. Also, I've get rid of unnecessary files creation, so the process should be much faster now.

@vmaxim Please take a look when you have time

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/handler/CaptureScreenshot.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/CaptureScreenshot.java
@@ -1,17 +1,44 @@
 package io.appium.uiautomator2.handler;
 
-import android.os.Environment;
+import android.app.UiAutomation;
+import android.graphics.Bitmap;
+import android.support.annotation.Nullable;
+import android.util.Base64;
 
-import java.io.File;
+import java.io.ByteArrayOutputStream;
 
 import io.appium.uiautomator2.handler.request.SafeRequestHandler;
 import io.appium.uiautomator2.http.AppiumResponse;
 import io.appium.uiautomator2.http.IHttpRequest;
+import io.appium.uiautomator2.model.internal.CustomUiDevice;
 import io.appium.uiautomator2.server.WDStatus;
-import io.appium.uiautomator2.utils.Device;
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/appium/uiautomator2/handler/GetElementScreenshot.java
```diff
@@ -42,27 +43,37 @@ public AppiumResponse safeHandle(IHttpRequest request) {
                 Logger.error("Element is not visible");
                 return new AppiumResponse(getSessionId(request), WDStatus.ELEMENT_NOT_VISIBLE);
             }
-            final Context context = InstrumentationRegistry.getInstrumentation().getTargetContext();
-            final File outputFile = File.createTempFile("screenshot", ".png",
-                    context.getCacheDir());
+            final Bitmap screenshot = uia.takeScreenshot();
```
vmaxim: Can we extract this dancing with screenshots into separate methods to do `handlers` more readable and avoid code duplication?

mykola-mokhnach: there is, actually, nothing much to extract. expect of this single line.

app/src/main/java/io/appium/uiautomator2/handler/CaptureScreenshot.java
```diff
@@ -1,17 +1,44 @@
...
+
+    @Nullable
+    private static String takeScreenshot() {
+        final Bitmap screenshot = uia.takeScreenshot();
+        if (screenshot == null) {
+            return null;
+        }
+        try {
+            final ByteArrayOutputStream stream = new ByteArrayOutputStream();
+            if (!screenshot.compress(Bitmap.CompressFormat.PNG, SCREENSHOT_COMPRESSION_QUALITY,
```
vmaxim: `PNG` is lossless format - quality will be ignored

mykola-mokhnach: true. Will set to 100

PR #117, opened 2018-01-23
expose data about last scroll

jlipps:
It is desirable to know certain things about scroll views / list views, for example how much content is in them in terms of pixels or child elements. Because elements don't exist until scrolled into view, this is not possible.

However, the scroll accessibility event does contain this info, so we can examine it retroactively after a scroll action.

This PR traps that data and puts it on the session object, so it can be retrieved by a call to get the session details.

PR #119, opened 2018-01-27
generalize contentSize's scrollableOffset

jlipps:
Turns out this wasn't working for anything but ScrollViews. We care about ListViews and GridViews as well. So we have to do some magic. 

This builds off #117 so its code will be visible here until that PR is merged.


app/src/main/java/io/appium/uiautomator2/core/UiObjectChildGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+            if (parent instanceof UiObject) {
+                try {
+                    child = ((UiObject) parent).getChild(new UiSelector().index(curChild));
+                } catch (UiObjectNotFoundException e) {
+                    child = null;
+                }
+            } else {
+                child = uiObject2Children.get(curChild);
+            }
+            curChild += 1;
```
mykola-mokhnach: or ++ curChild

app/src/main/java/io/appium/uiautomator2/core/UiObjectChildGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+                    childCount = 0;
+                }
+            } else {
+                uiObject2Children = ((UiObject2) parent).getChildren();
+                childCount = uiObject2Children.size();
+            }
+            curChild = 0;
+        }
+
+       @Override
```
mykola-mokhnach: mark as Nullable

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
```diff
@@ -79,70 +90,142 @@ public AppiumResponse safeHandle(IHttpRequest request) {
...
-                nodeInfo = AccessibilityNodeInfoGetter.fromUiObject(childObject);
-            } catch (UiAutomator2Exception ignored) {
+    private static int getScrollableOffsetByItemCount (AndroidElement uiScrollable, int itemCount) {
+        Logger.debug("Figuring out scrollableOffset via item count");
+        Object scrollObject = uiScrollable.getUiObject();
+
+        // here we loop through the children and get their bounds until the height differs, then
+        // regardless of whether we have a list or a grid, we'll know the height of an item/row
+        try {
+            int lastExaminedYInitVal = -999999; // something that could never be a 'top'
```
mykola-mokhnach: there is a built in Integer.MIN_VALUE constant for such purpose

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
```diff
@@ -79,70 +90,142 @@ public AppiumResponse safeHandle(IHttpRequest request) {
...
+            int lastExaminedItemY = lastExaminedYInitVal;
+            int numRowsExamined = 0;
+            int numRowsToExamine = 3; // examine a few rows since the top ones often have bad offsets
+            Object lastExaminedItem = null;
+
+            UiObjectChildGenerator gen = new UiObjectChildGenerator(scrollObject);
+            for (Object item : gen) {
+                Rect bounds = getElementBoundsInScreen(item);
+
+                if (item == null) {
```
mykola-mokhnach: should this check be placed before getElementBoundsInScreen ?

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
```diff
@@ -79,70 +90,142 @@ public AppiumResponse safeHandle(IHttpRequest request) {
...
+                    throw new UiObjectNotFoundException("Could not get child of scrollview");
+                }
+
+                itemsPerRow += 1;
```
mykola-mokhnach: ++itemsPerRow

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
```diff
@@ -79,70 +90,142 @@ public AppiumResponse safeHandle(IHttpRequest request) {
...
+        // now scroll a bit back and forth in the view to populate the lastScrollData we need
+        int x1 = bounds.centerX();
+        int y1 = bounds.centerY() + MINI_SWIPE_PIXELS;
+        int x2 = x1;
+        int y2 = y1 - (MINI_SWIPE_PIXELS * 2);
+        int tries = 0;
+        AccessibilityScrollData lastScrollData = null;
+
+        // generating scrolldata is flakey and doesn't always work, so try a number of times
+        while (tries < 10 && lastScrollData == null) {
```
vmaxim: Can you remove this while-loop and check the flakiness with  setting `enableNotificationListener=false`?

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
```diff
@@ -6,33 +6,44 @@
...
+
+import static io.appium.uiautomator2.utils.Device.getUiDevice;
 
 public class GetElementAttribute extends SafeRequestHandler {
 
+    // these constants are magic numbers experimentally determined to minimize flakiness in generating
+    // last scroll data used in getting the 'contentSize' attribute.
+    // TODO see whether anchoring these to time and screen size is more reliable across devices
+    private static int MINI_SWIPE_STEPS = 10;
+    private static int MINI_SWIPE_PIXELS = 600;
```
vmaxim: Why you name this "MINI"? It's a half of 1280x720 screen.

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
```diff
@@ -79,70 +90,142 @@ public AppiumResponse safeHandle(IHttpRequest request) {
...
-        ContentSize(Rect rect) {
-            width = rect.width();
-            height = rect.height();
-            top = rect.top;
-            left = rect.left;
+        // now scroll a bit back and forth in the view to populate the lastScrollData we need
+        int x1 = bounds.centerX();
+        int y1 = bounds.centerY() + MINI_SWIPE_PIXELS;
+        int x2 = x1;
+        int y2 = y1 - (MINI_SWIPE_PIXELS * 2);
```
vmaxim: Should we check that the calculated coords are still inside the bounds rect of element?

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
jlipps: yes, yes it should :-)

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
jlipps: ok, i will rebase so i get that code and check it out

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
jlipps: you're right. it was supposed to be MINI but i found with a small swipe the accessibility event was not generated. i think listview elements actually need to cross indexes in order to generate the event. but this could also be related to notification listener flakiness, so will experiment again with that turned off.

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
jlipps: yes, we should.

jlipps:
ok, rebased and all comments addressed. what do you think @vmaxim @mykola-mokhnach

app/src/main/java/io/appium/uiautomator2/core/EventRegister.java
```diff
@@ -15,39 +16,54 @@
...
-                    eventFilter, timeout);
-            Logger.debug("Retrieved accessibility event for scroll");
-        } catch (TimeoutException ignore) {
-            Logger.error("Expected to receive a scroll accessibility event but hit the timeout instead");
+        // actually run the swipe/scroll
+        runnable.run();
+
+        // wait a certain amount of time for any accessibility events generated to be caught by
+        // our callback, then stop watching for events
+        try { Thread.sleep(timeout); } catch (InterruptedException ign) {}
```
vmaxim: I really don't like this hard-coded sleep. What's wrong with `executeAndWaitForEvent`?
I can't see any possibility to get random event, because it [clear event queue before execute `Runnable`](
https://android.googlesource.com/platform/frameworks/base.git/+/master/core/java/android/app/UiAutomation.java#568).

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
```diff
@@ -106,16 +107,33 @@ private static int getScrollableOffset(AndroidElement uiScrollable) {
         int y1 = bounds.centerY() + MINI_SWIPE_PIXELS;
         int x2 = x1;
         int y2 = y1 - (MINI_SWIPE_PIXELS * 2);
+
+        // ensure that our xs and ys are within the bounds of the element
+        if (y1 > bounds.width()) {
+            y1 = bounds.width();
```
vmaxim: It's better to [avoid edges of element](https://android.googlesource.com/platform/frameworks/testing/+/master/uiautomator/library/core-src/com/android/uiautomator/core/UiScrollable.java#635).

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
```diff
@@ -79,70 +91,157 @@ public AppiumResponse safeHandle(IHttpRequest request) {
...
-            left = rect.left;
+        // now scroll a bit back and forth in the view to populate the lastScrollData we need
+        int x1 = bounds.centerX();
+        int y1 = bounds.centerY() + MINI_SWIPE_PIXELS;
+        int x2 = x1;
+        int y2 = y1 - (MINI_SWIPE_PIXELS * 2);
+
+        // ensure that our xs and ys are within the bounds of the element
+        if (y1 > bounds.width()) {
+            y1 = bounds.width();
```
vmaxim: It's better to [avoid edges of element](https://android.googlesource.com/platform/frameworks/testing/+/master/uiautomator/library/core-src/com/android/uiautomator/core/UiScrollable.java#635).

app/src/main/java/io/appium/uiautomator2/core/EventRegister.java
jlipps: Here's what I observed with `executeAndWaitForEvent`: running the same full-screen swipe many many times, the swipe executed perfectly every time. However, the event retrieved by `executeAndWaitForEvent` had a different `scrollY` every time too. Sometimes it was very small, meaning the captured event happened at the very beginning of the swipe. In other words, I was not reliably getting the information I needed. This solution at least does a lot better job. I agree it's horribly hacky; maybe you can help me find a better way to be sure we're getting the last accessibility event and not one from mid-scroll.

app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
jlipps: yeah, good point. we have access to the touchPadding i think anyway.

app/src/main/java/io/appium/uiautomator2/core/EventRegister.java
vmaxim: You can try it [this way ](https://android.googlesource.com/platform/frameworks/testing/+/master/uiautomator/library/core-src/com/android/uiautomator/core/InteractionController.java#360)

imurchie:
@jlipps What is the status of this? Seems to be hanging.

jlipps:
definitely going to finish all my PRs up here soon

jlipps:
ok @imurchie @mykola-mokhnach @vmaxim finally had time to make and test the requested change. anything else from your end?

app/src/main/java/io/appium/uiautomator2/core/EventRegister.java
```diff
@@ -2,46 +2,94 @@
...
 
 import io.appium.uiautomator2.model.AccessibilityScrollData;
 import io.appium.uiautomator2.model.AppiumUiAutomatorDriver;
+import io.appium.uiautomator2.model.NotificationListener;
 import io.appium.uiautomator2.model.Session;
 import io.appium.uiautomator2.utils.Logger;
 
 public abstract class EventRegister {
 
+    public static int EVENT_COOLDOWN_MS = 750;
```
vmaxim: This timeout should be configurable, because it adds almost one second to all commands wrapped by `EventRegister.runAndRegisterScrollEvents` (like `touch[Up|Down|Move]`, `swipe`) if execution of this commands did not generate scroll event.

jlipps: @vmaxim configurable as a capability, probably?

vmaxim: Via settings  - it should be available at runtime. You can use buildin timeout [setScrollAcknowledgmentTimeout](https://developer.android.com/reference/android/support/test/uiautomator/Configurator.html#setScrollAcknowledgmentTimeout(long)).

jlipps: @vmaxim ok what do you think about this

jlipps: oh, just saw your comment n/m

jlipps: @vmaxim i'm not sure settings is currently set up to work well with this use case. we don't actually want to call `setScrollAcknowledgementTimeout` i don't think, because we have to explicitly pass a `timeout` value into `executeAndWaitForEvent`. So we will want to retrieve the value from the setting every time we use this. Currently `AbstractSetting` isn't set up to save values, nor is there any kind of singleton thing going on with the actual settings classes, to ensure we can get the value back out.

i could do all of that refactoring, or just keep with the capability for now. what do you think?

vmaxim: OK,  let's keep it in d-caps, so users which don't care about scroll data can set it to zero and disable all this stuff

jlipps: Sounds good. This should be a pretty niche feature. Does the code as it stands look good to you then?

jlipps:
@vmaxim any further comments then?

vmaxim:
@jlipps nope.

jlipps:
published in `1.3.0`

PR #124, opened 2018-03-23
Add system clipboard handlers

mykola-mokhnach:
this add clipboard handling primitives, so one can automate copypaste operations better. Only plain text clipboard format is supported, but the implementation allows to extend this.

@vmaxim Please take a look at this when you have time.

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/handler/GetClipboard.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/GetClipboard.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.uiautomator2.handler;
+
+import android.app.Instrumentation;
+import android.support.test.InstrumentationRegistry;
+import android.util.Base64;
+
+import org.json.JSONObject;
+
+import java.nio.charset.StandardCharsets;
+
+import io.appium.uiautomator2.handler.request.SafeRequestHandler;
+import io.appium.uiautomator2.http.AppiumResponse;
+import io.appium.uiautomator2.http.IHttpRequest;
+import io.appium.uiautomator2.server.WDStatus;
+import io.appium.uiautomator2.utils.ClipboardHelper;
+import io.appium.uiautomator2.utils.ClipboardHelper.ClipDataType;
+import io.appium.uiautomator2.utils.Logger;
+
+public class GetClipboard extends SafeRequestHandler {
+    private final Instrumentation mInstrumentation = InstrumentationRegistry.getInstrumentation();
+
+    private static String toBase64String(String s) {
+        return Base64.encodeToString(s.getBytes(StandardCharsets.UTF_8), Base64.DEFAULT);
+    }
+
+    public GetClipboard(String mappedUri) {
+        super(mappedUri);
+    }
+
+    @Override
+    public AppiumResponse safeHandle(IHttpRequest request) {
+        Logger.info("Get Clipboard command");
+        ClipDataType contentType = ClipDataType.PLAINTEXT;
+        try {
+            JSONObject payload = getPayload(request);
+            if (payload.has("contentType")) {
+                contentType = ClipDataType.valueOf(payload
+                        .getString("contentType")
+                        .toUpperCase());
+            }
+            switch (contentType) {
+                case PLAINTEXT:
+                    return new AppiumResponse(getSessionId(request), WDStatus.SUCCESS,
+                            toBase64String(new ClipboardHelper(mInstrumentation.getTargetContext())
+                                    .getTextData()));
+                default:
+                    throw new IllegalArgumentException();
+            }
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException(
+                    String.format("Only '%s' content types are supported. '%s' is given instead",
+                            ClipDataType.supportedDataTypes(),
+                            contentType));
+        } catch (Exception e) {
+            return new AppiumResponse(getSessionId(request), WDStatus.UNKNOWN_ERROR, e);
+        }
+    }
+}
--- a/app/src/main/java/io/appium/uiautomator2/handler/GetClipboard.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/GetClipboard.java
@@ -63,7 +63,7 @@ public class GetClipboard extends SafeRequestHandler {
                     throw new IllegalArgumentException();
             }
         } catch (IllegalArgumentException e) {
-            throw new IllegalArgumentException(
+            return new AppiumResponse(getSessionId(request), WDStatus.UNKNOWN_ERROR,
                     String.format("Only '%s' content types are supported. '%s' is given instead",
                             ClipDataType.supportedDataTypes(),
                             contentType));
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/appium/uiautomator2/handler/SetClipboard.java
```diff
@@ -0,0 +1,61 @@
...
+            if (payload.has("contentType")) {
+                contentType = ClipDataType.valueOf(payload
+                        .getString("contentType")
+                        .toUpperCase());
+            }
+            if (payload.has("label")) {
+                label = payload.getString("label");
+            }
+            switch (contentType) {
+                case TEXT: ClipboardHelper.setTextData(label, (String) content);
```
jlipps: since the user can set contentType to something else, should we throw if it's not supported?

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
```diff
@@ -116,6 +133,8 @@ private void registerPostHandler() {
         register(postHandler, new UpdateSettings("/wd/hub/session/:sessionId/appium/settings"));
         register(postHandler, new NetworkConnection("/wd/hub/session/:sessionId/network_connection"));
         register(postHandler, new ScrollToElement("/wd/hub/session/:sessionId/appium/element/:id/scroll_to/:elementId"));
+        register(postHandler, new GetClipboard("/wd/hub/session/:sessionId/appium/clipboard/get"));
```
jlipps: shouldn't getting be a getHandler?

app/src/main/java/io/appium/uiautomator2/handler/SetClipboard.java
mykola-mokhnach: it will raise above in `ClipDataType.valueOf`

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
mykola-mokhnach: @dpgraham asked the same question :) No, this is a template for the future. For example, if we decide to get the content of non-primary clipboard or to filter the content by its type.

app/src/main/java/io/appium/uiautomator2/handler/SetClipboard.java
jlipps: ok

app/src/main/java/io/appium/uiautomator2/utils/ClipboardHelper.java
```diff
@@ -0,0 +1,72 @@
...
+        if (cm == null) {
+            throw new ClipboardError("Cannot receive ClipboardManager instance from the system");
+        }
+        return cm;
+    }
+
+    @Nullable
+    public static String getTextData() {
+        final ClipboardManager cm = getManager();
+        final ClipData cd = cm.getPrimaryClip();
```
vmaxim: I'd rather check [hasPrimaryClip()](https://developer.android.com/reference/android/content/ClipboardManager.html#hasPrimaryClip()) before calling ```getPrimaryClip()```.

app/src/main/java/io/appium/uiautomator2/handler/GetClipboard.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public AppiumResponse safeHandle(IHttpRequest request) {
+        Logger.info("Get Clipboard command");
+        ClipDataType contentType = ClipDataType.PLAINTEXT;
+        try {
+            JSONObject payload = getPayload(request);
+            if (payload.has("contentType")) {
+                contentType = ClipDataType.valueOf(payload
```
vmaxim: Maybe it's better to catch ```IllegalArgumentException``` exception and return more user-friendly response?

app/src/main/java/io/appium/uiautomator2/handler/GetClipboard.java
```diff
@@ -0,0 +1,64 @@
...
+            }
+            switch (contentType) {
+                case PLAINTEXT:
+                    return new AppiumResponse(getSessionId(request), WDStatus.SUCCESS,
+                            toBase64String(ClipboardHelper.getTextData()));
+            }
+        } catch (Exception e) {
+            return new AppiumResponse(getSessionId(request), WDStatus.UNKNOWN_ERROR, e);
+        }
+        return new AppiumResponse(getSessionId(request), WDStatus.SUCCESS, "");
```
vmaxim: Instead of silent noop command we can add the ```default``` section to ```switch`` statement and throw exception, if context type is not supported yet.

app/src/main/java/io/appium/uiautomator2/handler/GetClipboard.java
```diff
@@ -0,0 +1,64 @@
...
+            JSONObject payload = getPayload(request);
+            if (payload.has("contentType")) {
+                contentType = ClipDataType.valueOf(payload
+                        .getString("contentType")
+                        .toUpperCase());
+            }
+            switch (contentType) {
+                case PLAINTEXT:
+                    return new AppiumResponse(getSessionId(request), WDStatus.SUCCESS,
+                            toBase64String(ClipboardHelper.getTextData()));
```
vmaxim: The result of ```ClipboardHelper.getTextData()```can be null and should be checked to prevent NPE exception.

app/src/main/java/io/appium/uiautomator2/utils/ClipboardHelper.java
mykola-mokhnach: done

app/src/main/java/io/appium/uiautomator2/handler/GetClipboard.java
```diff
@@ -0,0 +1,64 @@
...
+import io.appium.uiautomator2.http.AppiumResponse;
+import io.appium.uiautomator2.http.IHttpRequest;
+import io.appium.uiautomator2.server.WDStatus;
+import io.appium.uiautomator2.utils.ClipboardHelper;
+import io.appium.uiautomator2.utils.ClipboardHelper.ClipDataType;
+import io.appium.uiautomator2.utils.Logger;
+
+public class GetClipboard extends SafeRequestHandler {
+
+    private static String toBase64String(String s) {
```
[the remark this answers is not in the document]
mykola-mokhnach: why so?

app/src/main/java/io/appium/uiautomator2/handler/GetClipboard.java
mykola-mokhnach: done

app/src/main/java/io/appium/uiautomator2/handler/GetClipboard.java
mykola-mokhnach: it will never reach the `default` section, because we do throw exception if the contentType value is not one of enum items

app/src/main/java/io/appium/uiautomator2/handler/GetClipboard.java
mykola-mokhnach: fixed

app/src/main/java/io/appium/uiautomator2/handler/GetClipboard.java
vmaxim: Right, but enum can contain unimplemented content type. And with ```default``` section noop response can be removed.

mykola-mokhnach: ok, added default

app/src/test/java/io/appium/uiautomator2/utils/ClipboardHelperTests.java
```diff
@@ -0,0 +1,43 @@
...
+import org.robolectric.annotation.Config;
+import org.robolectric.shadows.ShadowApplication;
+
+import io.appium.uiautomator2.test.BuildConfig;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(RobolectricTestRunner.class)
+@Config(constants = BuildConfig.class)
+public class ClipboardHelperTests {
```
vmaxim: I think that unit tests should not live in the main code.

mykola-mokhnach: I've put them into the same place where W3CActions tests are already placed by you

vmaxim: Ops... Now i see.

PR #133, opened 2018-04-08
add get settings

KazuCocoa:
After finishing this PR, we can get settings reverting https://github.com/appium/appium-uiautomator2-driver/pull/158 

```ruby
[3] pry(#<AppiumLibCoreTest::Android::DeviceTest>)> @@driver.get_settings
=> {"actionAcknowledgmentTimeout"=>3000,
 "allowInvisibleElements"=>false,
 "ignoreUnimportantViews"=>false,
 "elementResponseAttributes"=>false,
 "enableNotificationListener"=>true,
 "keyInjectionDelay"=>0,
 "scrollAcknowledgmentTimeout"=>200,
 "shouldUseCompactResponses"=>true,
 "waitForIdleTimeout"=>10000,
 "waitForSelectorTimeout"=>10000}
[4] pry(#<AppiumLibCoreTest::Android::DeviceTest>)> @@driver.update_settings('ignoreUnimportantViews' => true)
=> true
[5] pry(#<AppiumLibCoreTest::Android::DeviceTest>)> @@driver.get_settings
=> {"actionAcknowledgmentTimeout"=>3000,
 "allowInvisibleElements"=>false,
 "ignoreUnimportantViews"=>true,
 "elementResponseAttributes"=>false,
 "enableNotificationListener"=>true,
 "keyInjectionDelay"=>0,
 "scrollAcknowledgmentTimeout"=>200,
 "shouldUseCompactResponses"=>true,
 "waitForIdleTimeout"=>10000,
 "waitForSelectorTimeout"=>10000}
```

## TODO
- [x] implement other `no settings` settings and make numbers to integer
- [x] implement test cases
- [x] clean up

PR #138, opened 2018-04-10
Refactoring

KazuCocoa:
https://github.com/KazuCocoa/appium-uiautomator2-server/pull/1/files

I got a PR from @vmaxim and this PR is it to appium/appium-uiautomator2-server .
(Sorry for my notice after merging https://github.com/appium/appium-uiautomator2-server/pull/133 )

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/model/Session.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/Session.java
@@ -8,10 +8,11 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
+import static io.appium.uiautomator2.model.settings.Settings.ELEMENT_RESPONSE_ATTRIBUTES;
+import static io.appium.uiautomator2.model.settings.Settings.SHOULD_USE_COMPACT_RESPONSES;
+
 public class Session {
     public static final String SEND_KEYS_TO_ELEMENT = "sendKeysToElement";
-    public static final String CAP_SHOULD_USE_COMPACT_RESPONSES = "shouldUseCompactResponses";
-    public static final String CAP_ELEMENT_RESPONSE_ATTRIBUTES = "elementResponseAttributes";
     private String sessionId;
     private ConcurrentMap<String, JSONObject> commandConfiguration;
     private KnownElements knownElements;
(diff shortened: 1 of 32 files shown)


app/src/main/java/io/appium/uiautomator2/model/Session.java
```diff
@@ -33,15 +34,17 @@ public String getSessionId() {
...
         }
         return shouldUseCompactResponses;
     }
 
     public static String[] getElementResponseAttributes() {
-        if (Session.capabilities.containsKey(CAP_ELEMENT_RESPONSE_ATTRIBUTES)) {
-            return Session.capabilities.get(CAP_ELEMENT_RESPONSE_ATTRIBUTES).toString().split(",");
+        if (Session.capabilities.containsKey(ELEMENT_RESPONSE_ATTRIBUTES.toString())) {
+            return Session.capabilities.get(ELEMENT_RESPONSE_ATTRIBUTES.toString()).toString().
+                    split(",");
```
mykola-mokhnach: in java the dot separator is usually the first character if wrapped to a new string:

```
.toString()
.split(",")
```

app/src/main/java/io/appium/uiautomator2/model/settings/AbstractSetting.java
```diff
@@ -22,35 +22,45 @@
...
     }
 
+    public abstract T getValue();
+
     protected abstract void apply(T value);
 
     private T convertValue(Object value) {
-        if (!valueType.isInstance(value)) {
+        try {
+            if (valueType == Long.class && value instanceof Number) {
```
mykola-mokhnach: I assume it is obvious than if the first condition here is true then the second one is true as well.

vmaxim: Can you explain your point in more detail? I can't find any correlation between these two conditions.

mykola-mokhnach: I mean any value of Long class is a number instance

vmaxim: ```valueType``` represents the type of setting, but ```value``` is the desired value. Therefore these variables can belong to different classes.

PR #141, opened 2018-04-28
[WIP] Add a possibility to use xpath 2.0 for lookup queries

mykola-mokhnach:
I'm still not quite sure about how to properly write integration tests for this feature. Perhaps you might have some fresh ideas @vmaxim ?


mykola-mokhnach:
https://stackoverflow.com/questions/24290839/saxon-he-on-android

Shortly said, this sux on Android, but works on desktop. Sad but true.

mykola-mokhnach:
this will add unnecessary overhead, since it will be necessary to perform at least two extra network exchanges between the server and the client - to receive the tree and then to return the lookup result. Do you think it worth it?

Also we'll need a mechanism to distinguish xpath 2.0 queries from xpath 1.0

PR #142, opened 2018-04-29
Refactor XPathFinder class

mykola-mokhnach:
I've noticed that the current implementation is very messy with implicit dependencies and conditions across the code. This refactoring tries to isolate such problems and improve the general readability.

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/model/XPathFinder.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/XPathFinder.java
@@ -33,10 +33,11 @@ import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.xpath.XPath;
 import javax.xml.xpath.XPathConstants;
+import javax.xml.xpath.XPathExpression;
 import javax.xml.xpath.XPathExpressionException;
 import javax.xml.xpath.XPathFactory;
 
-import io.appium.uiautomator2.common.exceptions.ElementNotFoundException;
+import io.appium.uiautomator2.common.exceptions.InvalidSelectorException;
 import io.appium.uiautomator2.common.exceptions.UiAutomator2Exception;
 import io.appium.uiautomator2.core.UiAutomatorBridge;
 import io.appium.uiautomator2.utils.Attribute;
@@ -76,14 +77,14 @@ public class XPathFinder implements Finder {
         final Element domNode = toDOMElement((UiElement<?, ?>) context, document, uiElementsMapping);
         document.appendChild(domNode);
         final NodeList nodes;
+        final NodeInfoList matchesList = new NodeInfoList();
         try {
             nodes = (NodeList) XPATH_COMPILER
                     .compile(xPathString)
                     .evaluate(domNode, XPathConstants.NODESET);
         } catch (XPathExpressionException e) {
-            throw new ElementNotFoundException(e);
+            throw new InvalidSelectorException(e);
         }
-        final NodeInfoList matchesList = new NodeInfoList();
         for (int i = 0; i < nodes.getLength(); i++) {
             if (nodes.item(i).getNodeType() != Node.ELEMENT_NODE) {
                 continue;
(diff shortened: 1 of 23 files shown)


app/src/main/java/io/appium/uiautomator2/model/XPathFinder.java
```diff
@@ -51,235 +51,197 @@
...
-    try {
-      Field localName = element.getClass().getDeclaredField("localName");
-      localName.setAccessible(true);
-      localName.set(element, tag(className));
-    } catch (NoSuchFieldException e) {
-      Logger.error("Unable to set field localName:"+e.getMessage());
-    } catch (IllegalAccessException e) {
-      Logger.error("Unable to set field localName:"+e.getMessage());
+        for (UiElement<?, ?> child : uiElement.getChildren()) {
+            element.appendChild(toDOMElement(child, document));
```
vmaxim: Could we get rid of the recursion here to avoid stack overflow exception?

app/src/main/java/io/appium/uiautomator2/model/XPathFinder.java
```diff
@@ -51,235 +51,197 @@
...
-      }
     }
-  }
 
-  public static NodeInfoList getNodesList(String xpathExpression,  AccessibilityNodeInfo nodeInfo) throws InvalidSelectorException, ParserConfigurationException, UiAutomator2Exception {
-    if(nodeInfo == null) {
-      XPathFinder.refreshUiElementTree();
-    } else {
-      XPathFinder.refreshUiElementTree(nodeInfo);
+    public static NodeInfoList getNodesList(String xpathExpression,
```
vmaxim: It seems that static modifiers are unnecessary here and below.

mykola-mokhnach: with static is better from performance perspective than without, since such method calls are not going through vtable.

app/src/main/java/io/appium/uiautomator2/model/XPathFinder.java
mykola-mokhnach: do you think it is going to be a problem?

app/src/main/java/io/appium/uiautomator2/model/XPathFinder.java
vmaxim: It's not true nowadays.
https://ideone.com/goZurZ

app/src/main/java/io/appium/uiautomator2/model/XPathFinder.java
mykola-mokhnach: I'm afraid, that UiElement's interface does not allow to use non-recursive approach :(

app/src/main/java/io/appium/uiautomator2/model/XPathFinder.java
mykola-mokhnach: yep, the difference is very tiny. However, it still helps me to easily distinguish "utility" methods from these, that use class properties.

app/src/main/java/io/appium/uiautomator2/model/XPathFinder.java
```diff
@@ -51,235 +49,200 @@
...
+        final SparseArray<UiElement<?, ?>> uiElementsMapping = new SparseArray<>();
+        final Element domNode = toDOMElement((UiElement<?, ?>) context, document, uiElementsMapping);
+        document.appendChild(domNode);
+        final NodeList nodes;
+        try {
+            nodes = (NodeList) XPATH_COMPILER
+                    .compile(xPathString)
+                    .evaluate(domNode, XPathConstants.NODESET);
+        } catch (XPathExpressionException e) {
+            throw new ElementNotFoundException(e);
```
vmaxim: Maybe it would be better to throw  InvalidSelectorException so the handlers can return response with status WDStatus.INVALID_SELECTOR instead of NO_SUCH_ELEMENT.

mykola-mokhnach: changed. also configured all custom exceptions to be inherited from UiAutomator2Exception

PR #143, opened 2018-04-29
Add an endpoint for getting battery information

mykola-mokhnach:
just because this was requested

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/handler/GetBatteryInfo.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/GetBatteryInfo.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.uiautomator2.handler;
+
+import android.app.Instrumentation;
+import android.support.test.InstrumentationRegistry;
+
+import org.json.JSONObject;
+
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/appium/uiautomator2/utils/BatteryHelper.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * @return Current battery level as double value in range [0.0. 1.0],
+     * where 1.0 means 100% charge.
+     */
+    public double getLevel() {
+        final Intent batteryStatus = getBatteryStatusIntent();
+        final int level = batteryStatus.getIntExtra(BatteryManager.EXTRA_LEVEL, -1);
+        final int scale = batteryStatus.getIntExtra(BatteryManager.EXTRA_SCALE, -1);
+        return level * 1.0 / scale;
```
vmaxim: -1 * 1.0 / -1 = 1.0

mykola-mokhnach: fixed

app/src/main/java/io/appium/uiautomator2/handler/GetBatteryInfo.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Override
+    public AppiumResponse safeHandle(IHttpRequest request) {
+        Logger.info("Get Battery Info command");
+        final JSONObject response = new JSONObject();
+        final BatteryHelper batteryHelper = new BatteryHelper(mInstrumentation
+                .getTargetContext());
+        try {
+            response.put("level", batteryHelper.getLevel());
+            response.put("status", batteryHelper.getStatus());
```
KazuCocoa: Can we change the response name of `status` to `state` like iOS? 
Since when I implemented it in Ruby client, I faced https://github.com/appium/ruby_lib_core/pull/85#discussion_r185284872 error and it failed to parse `status` as an error code of JsonWiredProtocol.

mykola-mokhnach: I just wanted to keep system-specific terminology. In Android this indicator is called STATUS

jlipps: if we can, it'd be nice to have cross-platform compatibility for the API. is that possible here?

mykola-mokhnach: ok, I'll change rename than on driver level

PR #149, opened 2018-05-03
Update server runner

mykola-mokhnach:
Fixed the updated `isServerStopped` method name and moved the file into different folder, so it is now parseable by Android Studio


mykola-mokhnach:
Why this file is not visible in the Studio then?

mykola-mokhnach:
so it does not detect the compilation error?

mykola-mokhnach:
This file was actually broken after we renamed `isStopServer` property in https://github.com/appium/appium-uiautomator2-server/pull/140/files

mykola-mokhnach:
I've moved it under different folder, thus package name stays the same. @vmaxim Would it be ok?

vmaxim:
My fault - refactoring does not affects all builds variants in studio :(

PR #150, opened 2018-05-04
Properly configure source sets for gradle and fix instrumented test location

mykola-mokhnach:
uia2 integration tests are now broken because of the recent change: https://travis-ci.org/appium/appium-uiautomator2-driver/builds/374725894

Please merge and publish this PR ASAP

@vmaxim @dpgraham FYI


app/build.gradle
```diff
@@ -61,6 +61,13 @@ android {
         exclude 'META-INF/maven/com.google.guava/guava/pom.properties'
         exclude 'META-INF/maven/com.google.guava/guava/pom.xml'
     }
+    sourceSets {
+        androidTest {
+            java.srcDirs = ['src/androidTestE2eTest/java',
+                            'src/androidTestServer/java',
+                            'src/test/java']
```
vmaxim: It's unit tests - not Instrumented. They should not be in androidTest section.

mykola-mokhnach: removed unit tests

app/build.gradle
```diff
@@ -61,6 +61,12 @@ android {
         exclude 'META-INF/maven/com.google.guava/guava/pom.properties'
         exclude 'META-INF/maven/com.google.guava/guava/pom.xml'
     }
+    sourceSets {
+        androidTest {
+            java.srcDirs = ['src/androidTestE2eTest/java',
+                            'src/androidTestServer/java']
```
vmaxim: ```AppiumUiAutomator2Server``` - it's a ```while(true){}``` test. I think it's a not good idea to include this test in common set.

mykola-mokhnach: do you know another way to show it in the sources list?

vmaxim: What do you mean by "sources list"?
There are two build variants in the project. First is the server, second is e2e tests for the server.

mykola-mokhnach: I mean this file is not shown by Android Studio if we use the default Gradle settings

mykola-mokhnach: scr

vmaxim: image

mykola-mokhnach: it is visible in the project tree, but is not included into Android sources. This is how it looks without that Gradle entry:

scr2

mykola-mokhnach: I';ve moved the class to main source set if androidTests is not the best place for it

vmaxim: image

PR #155, opened 2018-05-06
Fix build version in build.gradle

alexttransisland:
An attempt to fix https://github.com/appium/appium/issues/10648 after recent update of appium uiautomator2 version to 1.9.1


mykola-mokhnach:
Version change is the part of publishing process. It is not needed to create a separate PR for it.

Issue #77, opened 2017-09-15
SendKeyToElement does not throw exception if AndroidElement.setText() failed.


vmaxim:
#### logcat:
> 09-15 21:09:06.292  4571  4597 D appium  : Sending plain text to element: 00
> 09-15 21:09:06.294  4571  4597 W UiObject2: AccessibilityNodeInfo#performAction(ACTION_SET_TEXT) failed

#### uiautomator2:
[SendKeysToElement.safeHandle()](https://github.com/appium/appium-uiautomator2-server/blob/8eb809776b2a48435ec8574e28a553279124dd90/app/src/main/java/io/appium/uiautomator2/handler/SendKeysToElement.java#L83)

#### uiautomator:
[SetText.execute()](https://github.com/appium/appium-android-bootstrap/blob/master/bootstrap/src/io/appium/android/bootstrap/handler/SetText.java#L87)

PR #159, opened 2018-05-09
Verify result of set element text command

mykola-mokhnach:
Previously sendKeys/value command would pass even if the actual text has not been set. now this should not be the case anymore. Addresses https://github.com/appium/appium/issues/10662

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -19,7 +19,7 @@ android {
     buildToolsVersion '26.0.2'
     defaultConfig {
         applicationId "io.appium.uiautomator2"
-        minSdkVersion 19
+        minSdkVersion 21
         targetSdkVersion 24
         versionCode 1
         archivesBaseName = "appium-uiautomator2"
--- a/app/src/main/java/io/appium/uiautomator2/handler/SendKeysToElement.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/SendKeysToElement.java
@@ -1,3 +1,19 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package io.appium.uiautomator2.handler;
 
 import android.support.test.uiautomator.UiObjectNotFoundException;
@@ -5,6 +21,7 @@ import android.support.test.uiautomator.UiObjectNotFoundException;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import io.appium.uiautomator2.common.exceptions.InvalidElementStateException;
 import io.appium.uiautomator2.common.exceptions.UiAutomator2Exception;
 import io.appium.uiautomator2.handler.request.SafeRequestHandler;
 import io.appium.uiautomator2.http.AppiumResponse;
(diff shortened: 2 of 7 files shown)


app/build.gradle
```diff
@@ -19,7 +19,7 @@ android {
     buildToolsVersion '26.0.2'
     defaultConfig {
         applicationId "io.appium.uiautomator2"
-        minSdkVersion 19
+        minSdkVersion 21
```
jlipps: is this a change that will affect users?

app/src/main/java/io/appium/uiautomator2/handler/SendKeysToElement.java
```diff
@@ -77,7 +94,10 @@ protected AppiumResponse safeHandle(IHttpRequest request) throws JSONException,
         if (!replace && currText != null) {
             text = currText + text;
         }
-        element.setText(text, unicodeKeyboard);
+        if (!element.setText(text, unicodeKeyboard)) {
+            throw new InvalidElementStateException(String.format("Cannot set the element to '%s'. " +
+                    "Did you locate a correct one?", text));
```
jlipps: `Did you interact with the correct element?`

app/src/main/java/io/appium/uiautomator2/utils/ElementHelpers.java
```diff
@@ -169,13 +171,10 @@ public static void setText(final Object element, final String text, final boolea
...
-        if (element instanceof UiObject2) {
-            UiObject2.class.cast(element).setText(textToSend);
-        } else if (element instanceof UiObject) {
-            UiObject.class.cast(element).setText(textToSend);
-        } else {
-            throw new UiAutomator2Exception("Unknown object type: " + element.getClass().getName());
-        }
+        Bundle args = new Bundle();
+        args.putCharSequence(AccessibilityNodeInfo.ACTION_ARGUMENT_SET_TEXT_CHARSEQUENCE, textToSend);
+        return nodeInfo.performAction(AccessibilityNodeInfo.ACTION_SET_TEXT, args);
```
jlipps: interesting, so we're switching to `performAction` rather than using the uiobject methods. I'm OK with this as long as the behavior is the same as before!

mykola-mokhnach: this code is called for both cases if you check the downstream implantation if OS API version is above KitKat

app/build.gradle
mykola-mokhnach: We do check this in the driver code as well and throw an error if the API version is below 21

app/src/main/java/io/appium/uiautomator2/utils/ElementHelpers.java
vmaxim: It's ok for ```UiObject2```. But for ```UiObject``` we will get a problem here:
1. ```findAccessibilityNodeInfo``` returns null if element does not exist
2. ```setText``` waits for element with ```Configurator.getWaitForSelectorTimeout()``` (I'am ok if we skip this part)

https://github.com/appium/appium-uiautomator2-server/blob/665f99151b2ef2f658d7555c40493c7c5dfb659a/app/src/main/java/io/appium/uiautomator2/core/AccessibilityNodeInfoGetter.java#L33

https://android.googlesource.com/platform/frameworks/uiautomator/+/android-support-test/src/main/java/android/support/test/uiautomator/UiObject.java#184

vmaxim:
#77

app/src/main/java/io/appium/uiautomator2/handler/SendKeysToElement.java
```diff
@@ -77,7 +94,10 @@ protected AppiumResponse safeHandle(IHttpRequest request) throws JSONException,
         if (!replace && currText != null) {
             text = currText + text;
         }
-        element.setText(text, unicodeKeyboard);
+        if (!element.setText(text, unicodeKeyboard)) {
+            throw new InvalidElementStateException(String.format("Cannot set the element to '%s'. " +
```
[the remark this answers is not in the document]
mykola-mokhnach: it is not necessarily a text. Can be a progress bar value as well

app/src/main/java/io/appium/uiautomator2/utils/ElementHelpers.java
```diff
@@ -133,9 +132,11 @@ private static void putNullable(JSONObject jsonObject, String fieldName,
      * @param element         - target element
      * @param text            - desired text
      * @param unicodeKeyboard - true, if text should be encoded to unicode
+     * @return true if the text has been successfully set
      */
-    public static void setText(final Object element, final String text, final boolean unicodeKeyboard) throws UiObjectNotFoundException {
-        String textToSend = text;
+    public static boolean setText(final Object element, final String text, final boolean unicodeKeyboard) {
```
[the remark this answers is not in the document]
mykola-mokhnach: added

app/src/main/java/io/appium/uiautomator2/utils/ElementHelpers.java
mykola-mokhnach: added null verification

PR #160, opened 2018-05-10
Limit the depth of xml dump

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/10204

I've set the maximum depth to 70. Should be enough for the most of the cases, I think %)


mykola-mokhnach:
@vmaxim Can you please check the PR?

PR #161, opened 2018-05-14
Fix sending of key codes with W3C actions interface

mykola-mokhnach:
Key codes conversion on server side is a pain, since we don't actually know what the clients wants. It would be smarter to do this on the client side and only let the server to accept native codes.


mykola-mokhnach:
It is quite complicated to adjust Android native keys strategy to what is currently implemented in Selenium. I'm just trying to make it as flexible as possible. Also, I'd like to add some helper methods for sendKeys generation for Java client after this PR is merged

PR #163, opened 2018-05-16
Add meta key properties to keyUp/keyDown W3C events

mykola-mokhnach:
Added meta key modifiers to the W3C key events, so it would be easier to handle the meta state


mykola-mokhnach:
Some more adjustments are needed to be made here. Will reopen later

PR #165, opened 2018-05-18
Fix appium/appium#10707

vmaxim:
```UiAutomation.mOnAccessibilityEventListener``` is no longer accessible on Android P

```
W tomator2.serve: Accessing hidden field Landroid/app/UiAutomation;->mOnAccessibilityEventListener:Landroid/app/UiAutomation$OnAccessibilityEventListener; (blacklist, reflection)
```
https://android-developers.googleblog.com/2018/02/improving-stability-by-reducing-usage.html

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/core/UiAutomation.java
+++ b/app/src/main/java/io/appium/uiautomator2/core/UiAutomation.java
@@ -16,6 +16,9 @@
 package io.appium.uiautomator2.core;
 
 import android.app.UiAutomation.OnAccessibilityEventListener;
+import android.support.annotation.Nullable;
+
+import io.appium.uiautomator2.common.exceptions.UiAutomator2Exception;
 
 import static io.appium.uiautomator2.utils.ReflectionUtils.getField;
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: app/src/main/java/io/appium/uiautomator2/core/UiAutomation.java [hunk 1]
```diff
@@ -34,9 +37,15 @@ public static UiAutomation getInstance() {
...
     }
 
+    @Nullable
     public OnAccessibilityEventListener getOnAccessibilityEventListener() {
-        return (OnAccessibilityEventListener) getField(android.app.UiAutomation.class,
-                FIELD_ON_ACCESSIBILITY_EVENT_LISTENER, uiAutomation);
+        try {
+            return (OnAccessibilityEventListener) getField(android.app.UiAutomation.class,
+                    FIELD_ON_ACCESSIBILITY_EVENT_LISTENER, uiAutomation);
+        } catch (UiAutomator2Exception e) {
```
mykola-mokhnach: I'd rather catch Exception

app/src/main/java/io/appium/uiautomator2/model/NotificationListener.java
```diff
@@ -34,10 +34,11 @@
     private static final NotificationListener INSTANCE = new NotificationListener();
     private static final int TOAST_CLEAR_TIMEOUT = 3500;
 
+    private final UiAutomation uiAutomation;
     private List<CharSequence> toastMessage = new CopyOnWriteArrayList<>();
     private long recentToastTimestamp = currentTimeMillis();
     private OnAccessibilityEventListener originalListener = null;
-    private final UiAutomation uiAutomation;
+    private boolean isListening;
```
mykola-mokhnach: volatile

app/src/main/java/io/appium/uiautomator2/model/NotificationListener.java
```diff
@@ -67,6 +67,7 @@ public void stop() {
             return;
         }
         Logger.debug("Stopping toast notification listener.");
+        isListening = false;
         uiAutomation.setOnAccessibilityEventListener(originalListener);
```
mykola-mokhnach: is it ok to set the listener to null even if we cannot receive the original value?

vmaxim: We have no other way to remove our listener. ```Null``` is ok.
https://android.googlesource.com/platform/frameworks/base.git/+/master/core/java/android/app/UiAutomation.java#1139

PR #179, opened 2018-07-26
Fix exceptions output

mykola-mokhnach:
Previously it was printing the whole action JSON instead of the particular failing action item
Addresses https://github.com/appium/appium/issues/11111


mykola-mokhnach:
@jlipps 

In the original log there was no duration:

> dbug UiAutomator2       {
> dbug UiAutomator2         "type": "pointerMove",
> dbug UiAutomator2         "x": 241,
> dbug UiAutomator2         "y": 280.5
> dbug UiAutomator2       },

PR #182, opened 2018-08-06
ArithmeticException fix

alexander-kachechka:
Added check for 0 to avoid an ArithmeticException

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
@@ -164,6 +164,11 @@ public class GetElementAttribute extends SafeRequestHandler {
             }
             Logger.debug("Determined there were " + itemsPerRow + " items per row");
 
+            if (itemsPerRow == 0) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/appium/uiautomator2/handler/GetElementAttribute.java
```diff
@@ -164,6 +164,11 @@ private static int getScrollableOffsetByItemCount(AndroidElement uiScrollable, i
             }
             Logger.debug("Determined there were " + itemsPerRow + " items per row");
 
+            if (itemsPerRow == 0) {
+                // Need to exit. Other case we will get an ArithmeticException
+                return 0;
```
mykola-mokhnach: I'd assume the resulting number of rows will be `itemCount` in such case rather than zero

alexander-kachechka: This method should return offset value for provided number of rows. 
itemCount as a result would be wrong. Return zero as a result can avoid to failing the test

PR #191, opened 2018-09-23
Add platform version and display density to /info endpoint

mykola-mokhnach:
This will allow to speed up driver startup by getting the necessary information directly from the server instead of performing slow adb calls


mykola-mokhnach:
@jlipps I did test the stuff and the returned values were in sync with these returned by adb. You could also double check it on your side. I'll make the necessary fixes if needed while it is still in beta

PR #195, opened 2018-10-03
Sign apk with default appium certificate

dpgraham:
* Does this on 'build' so that we publish an APK that contains the Appium certificate
* This is so that the server doesn't have to run 'apksigner' during a session

### DIFF
--- a/scripts/sign-apk.js
+++ b/scripts/sign-apk.js
@@ -1,0 +1,15 @@
+
+import { androidHelpers } from 'appium-android-driver';
+import { asyncify } from 'asyncbox';
+import path from 'path';
+import packageJson from '../package.json';
+
+async function main () {
+  // Signs the APK with the default Appium Certificate
+  const adb = await androidHelpers.createADB({});
+  const pathToApk = path.resolve('apks', `appium-uiautomator2-server-v${packageJson.version}.apk`);
+  await adb.sign(pathToApk);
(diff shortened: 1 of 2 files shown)


scripts/sign-apk.js
```diff
@@ -0,0 +1,15 @@
...
+import { asyncify } from 'asyncbox';
+import path from 'path';
+import packageJson from '../package.json';
+
+async function main () {
+  // Signs the APK with the default Appium Certificate
+  const adb = await androidHelpers.createADB({});
+  const pathToApk = path.resolve('apks', `appium-uiautomator2-server-v${packageJson.version}.apk`);
+  await adb.sign(pathToApk);
+  await adb.executeApksigner(['verify', '--print-certs', pathToApk]);
```
mykola-mokhnach: this will fail if apksigner is not installed on the publisher's machine

mykola-mokhnach:
It might be that the customer wants to use a custom keystore for signing. Would this work in such case?

scripts/sign-apk.js
dpgraham: I'll take that out. It was only there to print the cert so that I could verify that it was working.

dpgraham:
@mykola-mokhnach I think it should. All this is doing is changing the default signature. Unless I misunderstand how this works.

PR #196, opened 2018-10-03
Deprecate the unused endpoint

mykola-mokhnach:
The endpoint is useless and does nothing. See https://github.com/appium/appium-uiautomator2-driver/pull/222 for more details:


app/src/main/java/io/appium/uiautomator2/handler/AppStrings.java
```diff
@@ -1,51 +1,22 @@
...
-            dataInput.close();
-
-            JSONObject appStrings = new JSONObject(new String(jsonBytes, "UTF-8"));
-            Logger.debug("json loading complete ");
-            return new AppiumResponse(getSessionId(request), WDStatus.SUCCESS, appStrings);
-        } catch (IOException e) {
-            Logger.error("Error loading json from " + filePath + " : " + e);
-            throw new UiAutomator2Exception(e);
-        }
+        return new AppiumResponse(getSessionId(request), WDStatus.SUCCESS, Collections.emptyMap());
```
dpgraham: If a user tries calling this endpoint, will uiautomator2-driver provide a deprecation notice?

mykola-mokhnach: What would be the best to do? Write a string to logcat?

dpgraham: Maybe a warning, `Logger.warn("The unlock endpoint is does nothing and will be removed in a future release")`.... just so the user knows it's not doing anything.

mykola-mokhnach: added

dpgraham: Oh, it's /strings... I thought it was unlock.... same thing I guess.

mykola-mokhnach:
Would you prefer to remove it completely and raise major module version instead

mykola-mokhnach:
@jlipps the handler has been completely deleted

PR #197, opened 2018-10-04
APK presign fix

dpgraham:
* Transpile script instead of using babel-node

### DIFF
--- a/lib/scripts/sign-apk.js
+++ b/lib/scripts/sign-apk.js
@@ -1,8 +1,7 @@
-
 import { androidHelpers } from 'appium-android-driver';
 import { asyncify } from 'asyncbox';
 import path from 'path';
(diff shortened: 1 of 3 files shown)


lib/scripts/sign-apk.js
```diff
@@ -1,8 +1,7 @@
-
 import { androidHelpers } from 'appium-android-driver';
 import { asyncify } from 'asyncbox';
 import path from 'path';
-import packageJson from '../package.json';
+import packageJson from '../../../package.json';
```
mykola-mokhnach: should find-root be used instead?

dpgraham: I think it's okay to leave it as-is. `find-root` is for the case where it's uncertain where package.json will be

PR #199, opened 2018-10-09
Get rid of the obsolete UnicodeEncoder helper

mykola-mokhnach:
Since API19 `setText` method of UiObject(2) natively supports Unicode, so there is no need for these obsolete workarounds for the minimum API level 21, which is set for UIA2


mykola-mokhnach:
@imurchie it always worked. As you can see there `needsEncoding ` was always returning false

PR #200, opened 2018-10-10
Refactor element attributes retrieval

mykola-mokhnach:
this makes attributes lookup case insensitive and also simplifies the downstream logic

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/model/UiObject2Element.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/UiObject2Element.java
@@ -171,10 +171,7 @@ public class UiObject2Element implements AndroidElement {
         if (result == null) {
             return null;
         }
-        if (!(result instanceof String)) {
-            return String.valueOf(result);
-        }
-        return (String) result;
+        return (result instanceof String) ? (String) result : String.valueOf(result);
     }
 
     @Override
--- a/app/src/main/java/io/appium/uiautomator2/model/UiObject2Element.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/UiObject2Element.java
@@ -23,6 +23,7 @@ import android.support.test.uiautomator.UiObject;
 import android.support.test.uiautomator.UiObject2;
 import android.support.test.uiautomator.UiObjectNotFoundException;
 import android.support.test.uiautomator.UiSelector;
+import android.util.Range;
 import android.view.accessibility.AccessibilityNodeInfo;
 import android.widget.Toast;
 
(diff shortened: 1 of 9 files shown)


app/src/main/java/io/appium/uiautomator2/model/UiObject2Element.java
```diff
@@ -98,77 +100,107 @@ public String getText() {
...
+        if (attr == null) {
+            throw generateNoAttributeException("null");
         }
-        throw generateNoAttributeException(attr);
-    }
 
-    public boolean getBoolAttribute(final String attr) throws UiAutomator2Exception {
-        switch (attr) {
+        final Object result;
+        switch (attr.toLowerCase()) {
```
imurchie: Other than the `"displayed"` attribute, this looks the same as in `UiObjectElement.java`. Is there no way to make this shared code? In fact much of this class seems the same as that one.

mykola-mokhnach: yep I was also thinking about that. Unfortunately UiObject and UiObject2 don't share any common interfaces. The only way here is to use reflection, which I'm not very happy with.

imurchie: But both the classes implement `AndroidElement`. Is there no way to have an abstract class that they both extend, which implements the common stuff?

mykola-mokhnach: this will require to create a separate getter for each attribute in the main abstract class and override it in both inherited classes, which is even more code

imurchie: Ugh. Ok.

app/src/main/java/io/appium/uiautomator2/model/UiObject2Element.java
```diff
@@ -98,77 +100,107 @@ public String getText() {
...
-                return nodeInfo != null && nodeInfo.isPassword();
+                result = nodeInfo != null && nodeInfo.isPassword();
+                break;
             default:
                 throw generateNoAttributeException(attr);
         }
+        if (result == null) {
+            return null;
+        }
+        return (result instanceof String) ? (String) result : String.valueOf(result);
```
jlipps: if `result` is already a string, why do we cast it to one?

jlipps: how can a boolean be an instance of a String?

mykola-mokhnach: it cannot. That is why we have Sttring.valueOf

jlipps: the part that i'm talking about is different. the conditional checks whether `result instanceof String`. if that is true, we return `(String) result`. I'm saying that we shouldn't need to cast `result` to a `String` in the case where we've already confirmed that it is a string.

jlipps: got it, thanks. makes you want stuff like Swift where once you do a check you can use the assumptions guaranteed by the check inside the block.

PR #203, opened 2018-10-15
Address another NPE when class name equals to null

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/11513#issuecomment-429995307


vlprus:
Thanks a lot guys for the quick fix.
Finding elements is not causing any issues

PR #209, opened 2018-10-26
Update motion events generation logic

mykola-mokhnach:
This should address the issue where incorrect number of pointer events was generated


mykola-mokhnach:
@jlipps Could you please check your test again with the recent commit? This might also fix issues with multiple pointers

PR #210, opened 2018-10-27
Update motion events generation logic

mykola-mokhnach:
It is the same content as in https://github.com/appium/appium-uiautomator2-server/pull/209, which has been non intentionally corrupted %)


app/src/test/java/io/appium/uiautomator2/utils/w3c/W3CActionsPreprocessingTests.java
```diff
@@ -175,7 +176,29 @@ public void verifyValidActionChainPreprocessingWithCancelChange() throws JSONExc
...
                 "{\"type\": \"pointerUp\", \"button\": 0}]" +
                 "} ]");
-        assertThat(preprocessActions(actionJson).toString(), is(equalTo(processedJson.toString())));
+        assertThat(actionsPreprocessor.preprocess(actionJson).toString(), is(equalTo(processedJson.toString())));
+    }
+
+    @Test(expected = ActionsParseException.class)
+    public void verifyChainPreprocessingFailsIfMultiplePointerTypesArePresent() throws JSONException {
+        final JSONArray actionJson = new JSONArray("[ {" +
+                "\"type\": \"pointer\"," +
```
imurchie: Ugh. Does Java have nothing like a here-doc?

mykola-mokhnach: unfortunately not. They've just added something like that into java10 spec

jlipps:
(see gist for logs: https://gist.github.com/jlipps/852c7c65c49c0070f84bd479fd6dbd72)

jlipps:
FWIW i went back to the 1.9.1 official release and it is working, however the slowness issue is still present. i think the time conversion for the pointer move is problematic somehow.

KazuCocoa:
When I ran `double_click`, I got an error. => https://gist.github.com/KazuCocoa/c59caeb832fd709fc429d534a42f1176
command: https://github.com/appium/ruby_lib_core/pull/157/files#diff-1419dcc55a00830b8adc2b23544e6702R88
Is this issue? (I haven't tested with previous version yet, though)

A double tap action combined with pause and pointer_down by own worked.
https://github.com/appium/ruby_lib_core/pull/157/files#diff-1419dcc55a00830b8adc2b23544e6702R75

Except for it, it worked in some test cases.

KazuCocoa:
1.9.1 didn't work for some actions...

It's a helpful hint for users to understand why it doesn't work, I think.
(We can imagine it needs some delay because of OS side restrictions/stabilities, but many users probably don't)

mykola-mokhnach:
@KazuCocoa Added the case above into the set of unit tests

mykola-mokhnach:
> 1.9.1 didn't work for some actions...

Of course it did not. This feature has never been tested properly. This is the to have this PR and fix as much problems as possible and also improve the quality of the code.

PR #213, opened 2018-10-29
Avoid refreshing accessibility twice while executing xpath queries

mykola-mokhnach:
The accessibility is refreshed anyway when xml page source is built


app/src/main/java/io/appium/uiautomator2/handler/FindElement.java
```diff
@@ -93,7 +93,10 @@ protected AppiumResponse safeHandle(IHttpRequest request) throws JSONException,
     @Nullable
     private Object findElement(By by) throws ClassNotFoundException, UiAutomator2Exception,
             UiObjectNotFoundException {
-        refreshRootAXNode();
+        if (!(by instanceof By.ByXPath)) {
+            // The accessibility is refreshed when xml page source is built
```
KazuCocoa: 👍

PR #215, opened 2018-10-30
Update tests for more visibility

imurchie:
Separate out the multiple tests within `findElementWithContextId` and `toastVerificationTest` so when something fails it can be pinpointed.

The toast tests will still fail because https://github.com/appium/appium-uiautomator2-server/commit/919d382161acde4852ff3b34e54ebd188e460811 broke them.

### DIFF
--- a/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/DeviceCommandsTest.java
+++ b/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/DeviceCommandsTest.java
@@ -244,7 +244,7 @@ public class DeviceCommandsTest extends BaseTest {
     }
 
     @Test
-    public void findElementWithContextId() throws JSONException {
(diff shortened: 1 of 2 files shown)


app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/DeviceCommandsTest.java
```diff
@@ -244,7 +244,7 @@ public void verify500HTTPStatusCode() {
     }
 
     @Test
-    public void findElementWithContextId() throws JSONException {
+    public void findElementWithContextId1() throws JSONException {
```
mykola-mokhnach: Can we have better names for tests instead of just 1,2,3...?

imurchie: Different? Yes. Better? Not really.

PR #212, opened 2018-10-29
Return empty string if null

imurchie:
In certain circumstances (such as Java 1.7, Android-21 in our `appium-uiautomator2-driver` Travis build (see https://travis-ci.org/appium/appium-uiautomator2-driver/jobs/447774226#L3630-L3637)) the framework returns `null`, and we pass that back, leading to failures where `""` is expected.
```
info HTTP --> GET /wd/hub/session/a31a2b91-2d5b-45e5-8b82-fbe8a5e13061/element/e0205769-e494-4f91-bcc8-76c1b2cdbb8b/text
info HTTP {}
info MJSONWP (a31a2b91) Driver proxy active, passing request on via HTTP proxy
dbug JSONWP Proxy Matched '/wd/hub/session/a31a2b91-2d5b-45e5-8b82-fbe8a5e13061/element/e0205769-e494-4f91-bcc8-76c1b2cdbb8b/text' to command name 'getText'
dbug JSONWP Proxy Proxying [GET /wd/hub/session/a31a2b91-2d5b-45e5-8b82-fbe8a5e13061/element/e0205769-e494-4f91-bcc8-76c1b2cdbb8b/text] to [GET http://localhost:8200/wd/hub/session/a6cdb841-e905-49ad-ae43-51a9ac773de4/element/e0205769-e494-4f91-bcc8-76c1b2cdbb8b/text] with body: {}
dbug JSONWP Proxy Got response with status 200: "{\"sessionId\":\"a6cdb841-e905-49ad-ae43-51a9ac773de4\",\"status\":0,\"value\":null}"
info JSONWP Proxy Replacing sessionId a6cdb841-e905-49ad-ae43-51a9ac773de4 with a31a2b91-2d5b-45e5-8b82-fbe8a5e13061
info HTTP <-- GET /wd/hub/session/a31a2b91-2d5b-45e5-8b82-fbe8a5e13061/element/e0205769-e494-4f91-bcc8-76c1b2cdbb8b/text 200 39
```

PR #218, opened 2018-10-31
Update getText to return an empty value instead of null and bounds/displayed calculation

mykola-mokhnach:
The PR does the same as https://github.com/appium/appium-uiautomator2-server/pull/212 plus a bit more

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/core/AccessibilityNodeInfoHelpers.java
+++ b/app/src/main/java/io/appium/uiautomator2/core/AccessibilityNodeInfoHelpers.java
@@ -15,39 +15,71 @@
  */
 package io.appium.uiautomator2.core;
 
+import android.annotation.SuppressLint;
+import android.annotation.TargetApi;
 import android.graphics.Rect;
+import android.os.Build;
 import android.os.Bundle;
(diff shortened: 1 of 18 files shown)


app/src/main/java/io/appium/uiautomator2/core/AccessibilityNodeInfoHelpers.java
```diff
@@ -0,0 +1,149 @@
...
+import io.appium.uiautomator2.utils.Logger;
+
+import static io.appium.uiautomator2.utils.Device.getUiDevice;
+import static io.appium.uiautomator2.utils.StringHelpers.charSequenceToNullableString;
+import static io.appium.uiautomator2.utils.StringHelpers.charSequenceToString;
+
+/**
+ * This class contains static helper methods to work with {@link AccessibilityNodeInfo}
+ */
+public class AccessibilityNodeInfoHelpers {
```
dpgraham: Did you only rename this file or did you rename it _and_ make changes.

mykola-mokhnach: I did both

PR #224, opened 2018-11-13
Switch to StAX (XMLPull) implementation for xml source building

mykola-mokhnach:
I observe that DOM parser, which is built in into Android is kind'a non-maintained and contains old and annoying bugs. The StAX implementation by XMLPull looks better from this perspective, but might potentially break backwards compatibility. My proposal would be to merge this PR after 1.10 is released and give it some time for stabilization.


mykola-mokhnach:
@aluedeke FYI

mykola-mokhnach:
@imurchie Would you like to try if this PR properly works with apps that contain Unicode chars in class names?

imurchie:
Trying it out now.

PR #232, opened 2018-12-14
Update components and migrate to androidx

mykola-mokhnach:
It is time to upgrade UIA2 to use the newest available Android SDK.


dpgraham:
Maybe at some point we should just port the UiAutomator2 driver into the Espresso Driver and have the ability to switch between Espresso context and UiAutomator2 context.

PR #242, opened 2019-01-10
make logging configurable

### LINKED ISSUE #241: Need a way to disable logging
Is there any way to disable logging other than commenting it out and creating my own build?

sshock:
the logging level is INFO by default, and can be overridden with setprop or
/data/local.prop file (changing the latter requires rebooting the device)

for details see
https://developer.android.com/reference/android/util/Log#isLoggable(java.lang.String,%20int)

fixes #241


sshock:
(used a force-push to fix the author)

Issue #98, opened 2017-12-05
is that a bug in app/src/main/java/io/appium/uiautomator2/handler/ScrollTo.java+49 with WDStatus.UNKNOWN_ERROR


JerryLiu0821:
# is that a bug in app/src/main/java/io/appium/uiautomator2/handler/ScrollTo.java+49 with WDStatus.UNKNOWN_ERROR

out of try{} catch() {}, `WDStatus` should be `SUCCESS`

PR #244, opened 2019-01-23
Implementation of the 'scroll' command

Anton659011:
This change fixes implementation of the ScrollTo command, so that now it supports different strategies to search for a GUI element (by accessibility id, by class name, and by using the Android uiautomator selectors). The fix also addresses issue #98. Please let me know if I missed something in the implementation, and/or more files need to be updated.

Command: **POST /session/:sessionId/touch/scroll**
Here are a few examples of the supported JSON objects in the body of the POST request:

Scroll to an element using its accessibility id:
```
{
  "params":
  {
    "strategy":"accessibility id",
    "selector":"MyTestApp.OptionsPage.Option1.Label"
  }
}
```

Scroll to an element using its class name:
```
{
  "params":
  {
    "strategy":"class name",
    "selector":"android.widget.TextView"
  }
}
```

Scroll to an element identified by an Android uiautomator selector - 'element's text contains given string':
```
{
  "params":
  {
    "strategy":"-android uiautomator",
    "selector":"new UiSelector().textContains(\"Radio Group\")"
  }
}
```

Scroll to an element identified by an Android uiautomator selector - 'element's content-desc is the given string':
```
{
  "params":
  {
    "strategy":"-android uiautomator",
    "selector":"new UiSelector().description(\"MyTestApp.OptionsPage.Option2.Label\")"
  }
}
```


KazuCocoa:
Can you add some test cases?
It might help to what format is acceptable for the command.
https://github.com/appium/appium-uiautomator2-server/blob/ce4b766cf702e9d0b8b2304b06a8fd00aaff8fd6/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/DeviceCommandsTest.java
https://github.com/appium/appium-uiautomator2-server/blob/ce4b766cf702e9d0b8b2304b06a8fd00aaff8fd6/app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/commands/DeviceCommands.java#L194

app/src/main/java/io/appium/uiautomator2/utils/Device.java
```diff
@@ -36,11 +36,13 @@ public static void wake() throws RemoteException {
...
 
-    public static void scrollTo(String scrollToString) throws UiObjectNotFoundException {
-        // TODO This logic needs to be changed according to the request body from the Driver
+    public static void scrollToElement(UiSelector selector) throws UiObjectNotFoundException {
         UiScrollable uiScrollable = new UiScrollable(new UiSelector().scrollable(true).instance(0));
-        uiScrollable.scrollIntoView(new UiSelector().descriptionContains(scrollToString).instance(0));
-        uiScrollable.scrollIntoView(new UiSelector().textContains(scrollToString).instance(0));
+        boolean result = uiScrollable.scrollIntoView(selector);
+
+        if(!result) {
```
mykola-mokhnach: `if (!uiScrollable.scrollIntoView(selector))`

Anton659011:
I knew I missed something... 
Sure, I will add/update the tests. At the moment, whenever I try to run any of the **androideTestE2eTest** tests, I see an error like the following:
```
Process finished with exit code 1
Class not found: "io.appium.uiautomator2.unittest.test.DeviceCommandsTest"Empty test suite.
```
The other unit tests, however, run OK and all pass:
image

My development environment:
```
Android Studio 3.3 for Windows 10 64-bit
Build #AI-182.5107.16.33.5199772, built on December 24, 2018
JRE: 1.8.0_152-release-1248-b01 amd64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
```
What could be the problem, and how can I fix it? 
Thanks!

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/BaseTest.java
```diff
@@ -101,4 +103,18 @@ protected void clickAndWaitForStaleness(String elementId) throws JSONException {
         click(elementId);
         waitForElementInvisibility(elementId);
     }
+
+    protected void navigateTo(String navigationPath) throws JSONException {
+        for(String item : navigationPath.split("/")) {
+            By by = By.androidUiAutomator("new UiSelector().text(\"" + item + "\")");
```
mykola-mokhnach: use String.format

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/BaseTest.java
```diff
@@ -101,4 +103,18 @@ protected void clickAndWaitForStaleness(String elementId) throws JSONException {
...
+            Response response;
+            while(true)
```
mykola-mokhnach: please set some timeout

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/BaseTest.java
```diff
@@ -101,4 +103,18 @@ protected void clickAndWaitForStaleness(String elementId) throws JSONException {
...
+            {
+                response = findElement(by);
+                if(response.isSuccessful()) break;
```
mykola-mokhnach: Please always put blocks into curly brackets

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/TestUtils.java
```diff
@@ -165,6 +165,8 @@ public static JSONObject convertByToJson(By by, String contextId) {
                 strategy = "id";
             } else if (by instanceof By.ByXPath) {
                 strategy = "xpath";
+            } else if (by instanceof By.ByText) {
```
mykola-mokhnach: do we need one more selector with the same target?

app/src/main/java/io/appium/uiautomator2/utils/Device.java
```diff
@@ -36,11 +36,20 @@ public static void wake() throws RemoteException {
...
     }
 
-    public static void scrollTo(String scrollToString) throws UiObjectNotFoundException {
-        // TODO This logic needs to be changed according to the request body from the Driver
+    public static void scrollToElement(UiSelector selector, int maxSwipes) throws UiObjectNotFoundException {
         UiScrollable uiScrollable = new UiScrollable(new UiSelector().scrollable(true).instance(0));
-        uiScrollable.scrollIntoView(new UiSelector().descriptionContains(scrollToString).instance(0));
-        uiScrollable.scrollIntoView(new UiSelector().textContains(scrollToString).instance(0));
+
+        if(uiScrollable.getClassName().equals(android.widget.HorizontalScrollView.class.getName())) {
```
mykola-mokhnach: Please format the code

mykola-mokhnach: can .getClassName() return null?

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/DeviceCommandsTest.java
```diff
@@ -22,6 +22,7 @@
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.junit.Ignore;
```
mykola-mokhnach: is it used?

app/src/main/java/io/appium/uiautomator2/utils/Device.java
Anton659011: 1. I assume, the problem with code-formatting was the line slightly longer than 100 characters. I'll fix this.
2. All UI elements seem to have their **className** property set, so it is highly unlikely that  method **getClassName()** will return null for **UiScrollable**. However, it won't hurt to add a check-for null here.

mykola-mokhnach: > I assume, the problem with code-formatting was the line slightly longer than 100 characters. I'll fix this.

This one plus a space after `if`. Usually it is just enough to perform Cmd+Shift+L in IDEA

> All UI elements seem to have their className property set, so it is highly unlikely that method getClassName() will return null for UiScrollable. However, it won't hurt to add a check-for null here.

Feel free to use Objects.equal, which guarantees null-safery

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/DeviceCommandsTest.java
Anton659011: The run time of the **scrollVeryLongListSuccessfully** test is about 8.5 minutes, which might fail automated build jobs that have timeout of 10 minutes. To verify this scenario, I suggest ignoring this test in automated builds by adding the @Ignore attribute to it, and only running the test locally.

mykola-mokhnach: I see that only `shouldShutdownServerOnPowerDisconnect ` test has this attribute. Also, it would be useful to put a comment with the explanation above.

app/src/main/java/io/appium/uiautomator2/model/By.java
```diff
@@ -66,6 +66,13 @@ public static By androidUiAutomator(String expression) {
         return new ByAndroidUiAutomator(expression);
     }
 
+    public static By text(final String text) {
+        if (text == null)
```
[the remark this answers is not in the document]
Anton659011: The implementation of the new method followed the pattern used in the similar methods of the class. Accept as is.

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/TestUtils.java
Anton659011: **By.ByText** looks into the **text** property of a UI element. None of the other selectors seem to do that. Accept as is.

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/BaseTest.java
Anton659011: Fixed as suggested.

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/BaseTest.java
Anton659011: The timeout logic has been added.

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/BaseTest.java
Anton659011: Noted. Fixed.

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/commands/DeviceCommands.java
```diff
@@ -185,25 +185,115 @@ public static Response getSettings() {
...
+     * The given maximum number of swipes will be used during the element search.
+     *
+     * @param text
+     * @param maxSwipes
+     * @return Response from UiAutomator2 server
+     * @throws JSONException
+     */
+    public static Response scrollToText(String text, int maxSwipes) throws JSONException {
+        String textEscaped = text.replace("\"", "\\\"");
+        String uiSelectorSpec = "new UiSelector().text(\"" + textEscaped + "\")";
```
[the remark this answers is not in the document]
Anton659011: Fixed as suggested.

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/DeviceCommandsTest.java
Anton659011: Explanation about the long run time has been added to test **scrollVeryLongListSuccessfully**. 
The **shouldShutdownServerOnPowerDisconnect** was added the @Ignore attribute for local runs only, and was checked in by mistake. It should not be in the production code. Removed.

app/src/main/java/io/appium/uiautomator2/handler/ScrollTo.java
```diff
@@ -24,20 +28,53 @@ public ScrollTo(String mappedUri) {
...
+        String maxSwipes = "$.params.maxSwipes";
+
+        String strategyString = JsonPath.compile(strategy).read(json);
+        String selectorString = JsonPath.compile(selector).read(json);
+
+        int maxSearchSwipes;
+        try {
+            maxSearchSwipes = JsonPath.compile(maxSwipes).read(json);
+        } catch (InvalidPathException e) {
+            maxSearchSwipes = 0;
```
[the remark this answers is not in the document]
Anton659011: I have added logging of the fact that a default value will be used for the **maxSwipes** parameter. I can add more detailed log message(s) if this is not enough. In such a case, please suggest the right log message format.

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/BaseTest.java
```diff
@@ -101,4 +103,29 @@ protected void clickAndWaitForStaleness(String elementId) throws JSONException {
...
+            long timeoutMs = 60000;
+            long endTimeMs = System.currentTimeMillis() + timeoutMs;
+            Response response;
+            while(true)
+            {
+                response = findElement(By.text(item));
+                if(response.isSuccessful()) {
+                    break;
+                }
+                else {
```
mykola-mokhnach: else is redundant

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/TestUtils.java
mykola-mokhnach: then it makes sense to simply add it to the next entry using `||`

app/src/main/java/io/appium/uiautomator2/handler/ScrollTo.java
```diff
@@ -24,20 +28,54 @@ public ScrollTo(String mappedUri) {
...
+        String maxSwipes = "$.params.maxSwipes";
+
+        String strategyString = JsonPath.compile(strategy).read(json);
+        String selectorString = JsonPath.compile(selector).read(json);
+
+        int maxSearchSwipes;
+        try {
+            maxSearchSwipes = JsonPath.compile(maxSwipes).read(json);
+        } catch (InvalidPathException e) {
+            Logger.info("Parameter 'maxSwipes' not specified. Using default value.");
```
mykola-mokhnach: The parameter ... is either not specified or mistyped. Using 0 as the default value.

app/src/main/java/io/appium/uiautomator2/handler/ScrollTo.java
```diff
@@ -24,20 +28,54 @@ public ScrollTo(String mappedUri) {
...
+            maxSearchSwipes = 0;
         }
-        scrollTo(scrollToString);
-        Logger.info("Scrolled to String : ", scrollToString);
-        return new AppiumResponse(getSessionId(request), WDStatus.UNKNOWN_ERROR, true);
+
+        By by = new NativeAndroidBySelector().pickFrom(strategyString, selectorString);
+
+        UiSelector uiselector;
+        if(by instanceof By.ByAccessibilityId) {
```
mykola-mokhnach: space after if

app/src/main/java/io/appium/uiautomator2/model/By.java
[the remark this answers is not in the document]
mykola-mokhnach: I agree that this is more question of a style, but please enclose the if block into curly brackets then

app/src/main/java/io/appium/uiautomator2/utils/Device.java
```diff
@@ -36,11 +36,23 @@ public static void wake() throws RemoteException {
...
-        // TODO This logic needs to be changed according to the request body from the Driver
+    public static void scrollToElement(UiSelector selector, int maxSwipes)
+            throws UiObjectNotFoundException {
         UiScrollable uiScrollable = new UiScrollable(new UiSelector().scrollable(true).instance(0));
-        uiScrollable.scrollIntoView(new UiSelector().descriptionContains(scrollToString).instance(0));
-        uiScrollable.scrollIntoView(new UiSelector().textContains(scrollToString).instance(0));
+        String uiScrollableClassName = uiScrollable.getClassName();
+        String hScrollViewClassName = android.widget.HorizontalScrollView.class.getName();
+
+        if(uiScrollableClassName != null && uiScrollableClassName.equals(hScrollViewClassName)) {
```
mykola-mokhnach: missing space after if

mykola-mokhnach:
It looks like `unittest.test.DeviceCommandsTest.shouldBeAbleToFindElementViaUiScrollableScrollTextIntoView` is not stable on CI

app/src/main/java/io/appium/uiautomator2/model/By.java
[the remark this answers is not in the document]
Anton659011: Fixed as suggested (also in other methods - to make the style consistent throughout the file).

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/internal/BaseTest.java
Anton659011: Redundancy has been removed.

app/src/main/java/io/appium/uiautomator2/handler/ScrollTo.java
Anton659011: Log message wording has been changed as suggested.

app/src/main/java/io/appium/uiautomator2/utils/Device.java
Anton659011: Fixed this one and another similar issue a few lines below.

app/src/main/java/io/appium/uiautomator2/handler/ScrollTo.java
Anton659011: Fixed now.

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/DeviceCommandsTest.java
```diff
@@ -469,26 +473,162 @@ public void toastVerificationTest4() throws JSONException {
...
-        assertFalse(by + " should not be found", response.isSuccessful());
-        scrollTo(scrollToText);
-        response = findElement(by);
-        // After Scroll Element was found
-        assertTrue(by + " should be found", response.isSuccessful());
+        // On devices with Android API Level 23 and earlier, the tab labels are displayed
+        // in all-caps, whereas their actual text is rather like 'Tab 13', 'Tab 26', etc.
+        // The arguments of methods findElement() and scrollToText() are case-sensitive,
+        // therefore we need to use different sets of inputs for different target devices.
+        int apiLevel = (int) getField(UiDevice.class, "API_LEVEL_ACTUAL", getUiDevice());
```
mykola-mokhnach: I assume there are easier ways to get the actual api level

mykola-mokhnach: https://github.com/appium/appium-uiautomator2-server/blob/20de6220d4d9b888892b8e148a1e16b6aa8b874d/app/src/main/java/io/appium/uiautomator2/utils/DeviceInfoHelper.java#L78

mykola-mokhnach:
The tests are still failing on Travis. Please note that Travis performance is not very good, since it can only run arm-based AVD, which requires to put longer timeouts or more retries for e2e tests in comparison to a local run.

I like the recent source fixes.

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/DeviceCommandsTest.java
Anton659011: Thank you for the suggestion. This method returns a string which might be not as easy to use as an integer. I'll see what could be done with that.

Anton659011:
Hmmm, the CI tests that fail were not touched by this change at all. 
Probably the new tests that I introduced, which happen to run right before the failing ones, put the target device in some bad state. I am looking into this.

Anton659011:
In the latest push I replaced calls to **navigateTo()** - a new method introduced in this pull request - with **startActivity()** - a method used by other tests that were not touched by this change.

mykola-mokhnach:
There is only one failing test. If it passes locally then I assume we might mark it as ignored for now and add a comment that it's unstable on CI.

Anton659011:
There are _two_ tests that fail, and the reason for the failures seem to be the small number of maximum search swipes set in one of the tests that runs before the failing ones. The number of search swipes is held in a _static_ property of the **UiScrollable** class. That is, whenever a non-default number of search swipes is used during the scroll, we must always restore the setting after the operation. Otherwise all subsequent scrolls will use the new value, which might not be good. I believe, this is exactly what happens in the failing CI tests. The latest update addresses this issue.

app/src/main/java/io/appium/uiautomator2/utils/Device.java
```diff
@@ -36,11 +36,30 @@ public static void wake() throws RemoteException {
...
+
+        if (java.util.Objects.equals(uiScrollableClassName, hScrollViewClassName)) {
+            uiScrollable.setAsHorizontalList();
+        }
+
+        if (maxSwipes > 0) {
+            uiScrollable.setMaxSearchSwipes(maxSwipes);
+        }
+
+        boolean scrollResult = uiScrollable.scrollIntoView(selector);
```
mykola-mokhnach: can this throw an exception? Should we put the next operator into `finally` block?

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/DeviceCommandsTest.java
Anton659011: The current implementation that uses integers seems to serve the purpose quite well. I would rather keep it this way.

app/src/main/java/io/appium/uiautomator2/utils/Device.java
Anton659011: Yes, this actually might be a good idea. 
The declaration of **UiScrollable.scrollIntoView(...)** says that it can throw a **UiObjectNotFoundException**. I'll update the logic of **Device.scrollToElement()** accordingly.

PR #246, opened 2019-02-11
Speed up xpath and source commands

mykola-mokhnach:
There is no need to rebuild the UI hierarchy for XML creation, since it is always gets rebuilt before on findElement(s)/source call (inside `refreshRootAXNode`)

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/model/UiAutomationElement.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/UiAutomationElement.java
@@ -100,18 +100,16 @@ public class UiAutomationElement extends UiElement<AccessibilityNodeInfo, UiAuto
         this.children = mutableChildren;
     }
 
-    public static UiAutomationElement rebuildForNewRoot(AccessibilityNodeInfo rawElement,
-                                                        @Nullable List<CharSequence> toastMSGs) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/appium/uiautomator2/core/AccessibilityNodeInfoDumper.java
```diff
@@ -165,10 +165,8 @@ private InputStream toStream() throws IOException {
             serializer.setOutput(outputStream, XML_ENCODING);
             serializer.startDocument(XML_ENCODING, true);
             serializer.setFeature("http://xmlpull.org/v1/doc/features.html#indent-output", true);
-            final UiElement<?, ?> xpathRoot = root == null
-                    ? rebuildForNewRoot(currentActiveWindowRoot(), NotificationListener.getInstance().getToastMessage())
-                    : rebuildForNewRoot(root, null);
-            serializeUiElement(xpathRoot, 0);
+            final UiElement<?, ?> xpathRoot = root == null ? getCachedRoot() : getCachedElement(root);
```
sravanmedarapu: Looks like we are ignoring ToastMessage?

mykola-mokhnach: this has been changed

PR #247, opened 2019-02-14
Avoid NPE on getting session details

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/12162

@jlipps There are several other places where getSession is called like this without checking for nullability. I'd better leave it to you to perform the necessary fixes, since the code has to do with AccessibilityScrollData stuff.


dpgraham:
This fix is in Appium 1.11.1-rc.0 (release candidate)

PR #262, opened 2019-04-05
[wip] Azure pipelines

dpgraham:


dpgraham:
@mykola-mokhnach would you, or somebody more familiar with this project, be able to help me with this? There's some test failures when running the JUnit tests that need to be resolved before I can push this.

mykola-mokhnach:
@dpgraham To me it looks like these toasts are not shown at all. Have you tried to run this test locally with the given emulator version? Maybe this happens because the test app is not optimised for this API or it's just a race condition.

Regarding the screenshot - is the test headless? If yes then the test should be skipped. This is what SkipHeadlessDevices annotation is for

PR #264, opened 2019-04-11
Use UIAutomator screenshoter as a fallback

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/12470

@kamza-tricentis FYI


mykola-mokhnach:
@KazuCocoa This is Travis emulator issue. Should be better after https://github.com/appium/appium-uiautomator2-server/pull/262 is merged

Issue #280, opened 2019-06-01
Refresh operation in getElementFromCache does not include the original element's context


kelly987654:
When interacting with an element that has gone stale, the ui-automator server attempts to reacquire the element using the element's original locator. However the query used to refresh the element does not include the context that may have been included with the element's original query. As a result, the re-acquired element found by the incomplete query may be a different element. This can result in the user unknowingly interacting with a different element than expected/originally acquired.

Code of interest is in KnownElements.java, getElementFromCache.

Example to reproduce (based on simple page layout below):
1. Get element by accessibility id: 'containerId'
2. Get element by className: android.widget.Button, including a context elementId acquired in step 1
3. Perform some operation so the elements on the page go stale (e.g. scroll page up and down)
4. Attempt to tap on the element acquired in Step 2 - The element is reacquired and matches on the topButton instead, resulting on a 'tap' on the topButton instead of the expected button.

```
<page>
  <topButton className='android.widget.Button'/>
  <container content-desc='containerId'>
    <buttonInContainer className='android.widget.Button'/>
  </container>
</page>
```

mykola-mokhnach:
The fix has been published in appium@beta

kelly987654:
Thank you!

PR #281, opened 2019-06-01
Fix cache restoration for elements with context

mykola-mokhnach:
Addresses https://github.com/appium/appium-uiautomator2-server/issues/280

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/handler/FirstVisibleView.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/FirstVisibleView.java
@@ -78,7 +78,7 @@ public class FirstVisibleView extends SafeRequestHandler {
         }
 
         String id = UUID.randomUUID().toString();
-        AndroidElement androidElement = getAndroidElement(id, firstObject, null);
+        AndroidElement androidElement = getAndroidElement(id, firstObject, false);
         session.getKnownElements().add(androidElement);
         JSONObject result = new JSONObject();
         result.put("ELEMENT", id);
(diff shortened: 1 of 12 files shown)


app/src/main/java/io/appium/uiautomator2/handler/FirstVisibleView.java
```diff
@@ -78,7 +78,7 @@ protected AppiumResponse safeHandle(IHttpRequest request) throws UiObjectNotFoun
         }
 
         String id = UUID.randomUUID().toString();
-        AndroidElement androidElement = getAndroidElement(id, firstObject, null);
+        AndroidElement androidElement = getAndroidElement(id, firstObject, false);
```
KazuCocoa: Should not here be `true`?

app/src/main/java/io/appium/uiautomator2/model/KnownElements.java
```diff
@@ -47,6 +47,67 @@ private String getCacheKey(AndroidElement element) {
...
     }
 
+    private void restoreCachedElement(AndroidElement element) {
+        final By by = element.getBy();
+        if (by == null) {
+            throw new StaleElementReferenceException(String.format(
+                    "The element '%s' does not exist in DOM anymore", element.getId()));
+        }
+
+        // We cannot restore a single element from a locator that matches multiple items
```
KazuCocoa: 👍

app/src/main/java/io/appium/uiautomator2/handler/FirstVisibleView.java
KazuCocoa: ah... visible element might be multiple. we cannot say it is an exactly single element, so false...?

mykola-mokhnach: exactly

PR #286, opened 2019-07-06
return timezone and locale as deviceInfo

KazuCocoa:
https://github.com/appium/appium-xcuitest-driver/pull/1003 for UIA2
For Espresso driver is https://github.com/appium/appium-espresso-driver/pull/465

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/utils/DeviceInfoHelper.java
+++ b/app/src/main/java/io/appium/uiautomator2/utils/DeviceInfoHelper.java
@@ -34,8 +34,11 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/appium/uiautomator2/utils/DeviceInfoHelper.java
```diff
@@ -261,4 +264,26 @@ public String getRealDisplaySize() {
...
+    /**
+     * Get current system timezone
+     * e.g. "Asia/Tokyo", "America/Caracas"
+     *
+     */
+    @NonNull
+    public String getTimeZone() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            return TimeZone.getDefault().toZoneId().getId();
+        } else {
```
mykola-mokhnach: else is redundant

PR #289, opened 2019-07-15
feat: do not suppress accessibility services

KazuCocoa:
Related to https://github.com/appium/appium/issues/4910

Set `UiAutomation.FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES` if `DISABLE_SUPPRESS_ACCESSIBILITY_SERVICES`key and value, `true` or `false` is given as a parameter of instrument.

Respect existing setting if no `DISABLE_SUPPRESS_ACCESSIBILITY_SERVICES` key is given


KazuCocoa:
Do you have time to run this change with the accessibility service enabled app?

I haven't tested this case, https://github.com/appium/appium/issues/4910, well since I had no such app.
If someone could test this change and work fine, I'd like to continue to work this and https://github.com/appium/appium-espresso-driver/pull/468

tryge:
We can run the test on Monday, what do we need to do? (We've only ever used appium releases so far...).

KazuCocoa:
okay, thanks.
I will leave a comment on how to use this apk in your release env later.

KazuCocoa:
1. Download [apks.zip](https://github.com/appium/appium-uiautomator2-server/files/3774905/apks.zip)
2. Unzip and copy them into `/usr/local/lib/node_modules/appium/node_modules/appium-uiautomator2-server/apks` on your local
    - Overwrite: appium-uiautomator2-server-debug-androidTest.apk
    - Rename _appium-uiautomator2-server-v4.3.0.apk_ in the zip to your local machine's _appium-uiautomator2-server-vX_X_X.apk_ and overwrite it.
        - The name depends on appium version
3. Uninstall UIA2 servers from your test device
4. Run your tests

Then, make sure the above two overwritten apks will be installed on your Appium log.

tryge:
We tried to get it to work with these apks, but we haven't managed to do so. What do you need from us?

KazuCocoa:
The result of whether this UIA2 server apk works with the app under test which uses accessibility service without stopping the accessibility service

tryge:
We didn't see a difference in behaviour.

KazuCocoa:
thanks :)

app/src/main/java/io/appium/uiautomator2/utils/Device.java
```diff
@@ -22,6 +25,14 @@
 public abstract class Device {
 
     public static UiDevice getUiDevice() {
+        boolean disableSuppressAccessibilityService = Boolean.parseBoolean(
+                InstrumentationRegistry.getArguments().getString("DISABLE_SUPPRESS_ACCESSIBILITY_SERVICES", "false"));
+        if (disableSuppressAccessibilityService && Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            // The flag is necessary not to stop running accessibility service
+            // https://github.com/appium/appium/issues/4910
+            // https://developer.android.com/reference/android/app/UiAutomation.html#FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES
+            Configurator.getInstance().setUiAutomationFlags(UiAutomation.FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES);
```
mykola-mokhnach: should we do this each time we get the instrumentation or it would be enough to call the configurator once?

KazuCocoa:
I ensured this logic also works well with `Configurator.getInstance().getUiAutomationFlags()` in `getUiDevice` by breakpoint debugging.

`Configurator.getInstance().getUiAutomationFlags()` is `0` without `UiAutomation.FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES` via `disableSuppressAccessibilityService: false` in https://github.com/appium/appium-uiautomator2-driver/pull/376
`Configurator.getInstance().getUiAutomationFlags()` is `1` with `UiAutomation.FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES` via `disableSuppressAccessibilityService: true` in https://github.com/appium/appium-uiautomator2-driver/pull/376

app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
```diff
@@ -50,13 +56,33 @@
...
 
+    private static boolean shouldDisableSuppressAccessibilityService = Boolean.parseBoolean(
+            InstrumentationRegistry.getArguments().getString("DISABLE_SUPPRESS_ACCESSIBILITY_SERVICES", "false"));
+
+    private void setDefaultAutomationFlag() {
+        // Need to ensure enabling suppress accessibility service
+        Configurator.getInstance().setUiAutomationFlags(0);
+    }
+
+    private void disableSuppressAccessibilityService() {
```
mykola-mokhnach: I'd rather rename this method to setAccessibilityServiceState

app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
```diff
@@ -50,13 +56,33 @@
...
+
     private ServerInstrumentation(Context context, int serverPort) {
         if (!isValidPort(serverPort)) {
             throw new UiAutomator2Exception(String.format(
                     "The port is out of valid range [%s;%s]: %s", MIN_PORT, MAX_PORT, serverPort));
         }
         this.serverPort = serverPort;
         this.powerManager = (PowerManager) context.getSystemService(Context.POWER_SERVICE);
+
+        setDefaultAutomationFlag();
```
mykola-mokhnach: should we always reset these flags or only before we actually change them?

KazuCocoa: Good point.
I continued to observe the lifecycle of `Configurator.getInstance()` after the latest commit.
The value could remain before/after uia2 server lifecycle.

So, changing the state explicitly is better.

KazuCocoa:
Updated

PR #293, opened 2019-08-07
Make some caps configurable in runtime

mykola-mokhnach:
It does not make much sense to keep these settings from being changed in runtime

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/model/settings/ShouldUseCompactResponses.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/settings/ShouldUseCompactResponses.java
@@ -41,5 +41,5 @@ public class ShouldUseCompactResponses extends AbstractSetting<Boolean> {
                 .getSessionOrThrow()
                 .setCapability(SETTING_NAME, shouldUseCompactResponses);
     }
-    
+
 }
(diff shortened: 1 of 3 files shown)


imurchie:
My only thought is that if the caps/settings were named in the commit it would be _much_ easier for the release process.

mykola-mokhnach:
Ok, will enumerate cap names while squashing the PR

app/src/main/java/io/appium/uiautomator2/model/settings/ShouldUseCompactResponses.java
```diff
@@ -37,7 +36,10 @@ public Boolean getValue() {
...
     @Override
     protected void apply(Boolean shouldUseCompactResponses) {
-        Logger.debug("Dummy setting. Maintained in Session.capabilities.");
+        AppiumUIA2Driver
+                .getInstance()
+                .getSessionOrThrow()
+                .setCapability(SETTING_NAME, shouldUseCompactResponses);
     }
-
+
```
KazuCocoa: extra spaces?

mykola-mokhnach: it's idea formatter, perhaps a bug there %/

PR #294, opened 2019-08-15
feat: Switch UiAutomator2 server to W3C protocol

mykola-mokhnach:
This change is breaking and won't allow to init the server using JSONWP caps anymore.

Affected areas:
- Session capabilities 
- Element key in responses
- Errors in responses


app/src/test/java/io/appium/uiautomator2/utils/w3c/W3CCapsUtilsTests.java
```diff
@@ -0,0 +1,115 @@
...
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.junit.Test;
+
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+
+public class W3CCapsUtilsTests {
```
KazuCocoa: How about adding a test case which has non-appium prefix caps like `"{\"goog:chromeOptions\":{\"args\":[\"--disable-popup-blocking\"]}}"`? It should be valid.

mykola-mokhnach: yep, will do. In general, such caps should stay unchanged

app/src/main/java/io/appium/uiautomator2/utils/JSONUtils.java
```diff
@@ -57,4 +64,44 @@ public static Integer readInteger(JSONObject payload, String name, boolean isReq
         }
         return null;
     }
+
+    public static Map<String, Object> toMap(JSONObject obj) {
```
KazuCocoa: Nullable annotation?

app/src/main/java/io/appium/uiautomator2/utils/JSONUtils.java
```diff
@@ -57,4 +64,44 @@ public static Integer readInteger(JSONObject payload, String name, boolean isReq
...
+            } else if (value instanceof JSONArray) {
+                result.put(key, toList((JSONArray) value));
+            } else {
+                result.put(key, value);
+            }
+        }
+        return result;
+    }
+
+    public static List<Object> toList(JSONArray obj) {
```
KazuCocoa: Nullable?

app/src/main/java/io/appium/uiautomator2/utils/JSONUtils.java
mykola-mokhnach: the result will never be null. It can be an empty map in the worst case. Same below

KazuCocoa: Ah, okay. I thought null since they returned null when obj == null

PR #298, opened 2019-08-23
feat: Add server setting which allows to control the global wake lock

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/13111

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
+++ b/app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
@@ -21,8 +21,8 @@ import android.content.Context;
 import android.content.Intent;
 import android.os.Looper;
 import android.os.PowerManager;
-import android.os.RemoteException;
 
+import android.os.SystemClock;
 import io.appium.uiautomator2.common.exceptions.SessionRemovedException;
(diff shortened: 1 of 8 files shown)


app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
```diff
@@ -73,21 +75,40 @@ private void releaseWakeLock() {
...
+        return (wakeLock == null || !wakeLock.isHeld() || wakeLockAcquireTimestampMs <= 0 || wakeLockTimeoutMs <= 0)
+                ? 0
+                : wakeLockAcquireTimestampMs + wakeLockTimeoutMs - SystemClock.elapsedRealtime();
+    }
+
+    public void acquireWakeLock(long msTimeout) {
         releaseWakeLock();
 
+        if (msTimeout <= 0) {
+            return;
```
KazuCocoa: How about `Logger.warn` here or `.info` before this control flow to leave a message what number the user tried to set? It helps to track the configurable number.

mykola-mokhnach: makes sense, will add

PR #301, opened 2019-09-11
Improve support for multi windows and add capability "enableMultiWindows"

Volune:
## Motivation

This PR is initially to resolve an issue with `android.widget.PopupWindow`, where appium cannot interact with a PopupWindow which is not *focusable* (because it is not made the current window).
Also see [UIAutomator not able to access android.widget.PopupWindow](https://issuetracker.google.com/issues/37017411)

## Changes

- Use `getWindowRoots()` instead of `currentActiveWindowRoot()` wherever possible
- Add capability `enableMultiWindows` to enable multi windows. It was previously hardcoded as disabled because of [UiAutomatorViewer should Support multi-window to make it compatible with UiAutomator V2](https://issuetracker.google.com/issues/37095166)
- The capability is `false` by default, so no behavior change by default.

## Tests

- No new automated tests (advice welcome if needed)
- Ran existing tests
- Manually tested with appium desktop and option enabled/disabled

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/model/UiAutomationElement.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/UiAutomationElement.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.WeakHashMap;
 
 import androidx.annotation.Nullable;
+
 import io.appium.uiautomator2.core.AccessibilityNodeInfoHelpers;
 import io.appium.uiautomator2.utils.Attribute;
 import io.appium.uiautomator2.utils.Logger;
@@ -92,15 +93,17 @@ public class UiAutomationElement extends UiElement<AccessibilityNodeInfo, UiAuto
         this.children = buildChildren(node);
     }
 
-    protected UiAutomationElement(String hierarchyClassName, AccessibilityNodeInfo childNode, int index) {
+    protected UiAutomationElement(String hierarchyClassName, AccessibilityNodeInfo[] childNodes, int index) {
         super(null);
         Map<Attribute, Object> attribs = new LinkedHashMap<>();
         put(attribs, Attribute.INDEX, index);
         put(attribs, Attribute.CLASS, hierarchyClassName);
         this.attributes = Collections.unmodifiableMap(attribs);
-        List<UiAutomationElement> mutableChildren = new ArrayList<>();
-        mutableChildren.add(new UiAutomationElement(childNode, 0));
-        this.children = mutableChildren;
+        List<UiAutomationElement> children = new ArrayList<>();
+        for (AccessibilityNodeInfo childNode : childNodes) {
+            children.add(new UiAutomationElement(childNode, children.size()));
+        }
+        this.children = children;
     }
 
     private int getDepth() {
@@ -111,9 +114,9 @@ public class UiAutomationElement extends UiElement<AccessibilityNodeInfo, UiAuto
         this.depth = depth;
     }
 
-    public static UiAutomationElement rebuildForNewRoot(AccessibilityNodeInfo rawElement, @Nullable List<CharSequence> toastMSGs) {
+    public static UiAutomationElement rebuildForNewRoot(AccessibilityNodeInfo[] rawElements, @Nullable List<CharSequence> toastMSGs) {
         cache.clear();
-        UiAutomationElement root = new UiAutomationElement(ROOT_NODE_NAME, rawElement, 0);
+        UiAutomationElement root = new UiAutomationElement(ROOT_NODE_NAME, rawElements, 0);
         if (toastMSGs != null && !toastMSGs.isEmpty()) {
             for (CharSequence toastMSG : toastMSGs) {
                 Logger.debug("Adding toastMSG to root:" + toastMSG);
@@ -124,11 +127,7 @@ public class UiAutomationElement extends UiElement<AccessibilityNodeInfo, UiAuto
     }
 
     @Nullable
-    public static UiAutomationElement getCachedElement(AccessibilityNodeInfo rawElement,
-                                                       AccessibilityNodeInfo windowRoot) {
-        if (cache.get(rawElement) == null) {
-            rebuildForNewRoot(windowRoot, null);
-        }
+    public static UiAutomationElement getCachedElement(AccessibilityNodeInfo rawElement) {
         return cache.get(rawElement);
     }
 
--- a/app/src/main/java/io/appium/uiautomator2/model/settings/EnableMultiWindows.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/settings/EnableMultiWindows.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.uiautomator2.model.settings;
+
+import io.appium.uiautomator2.model.AppiumUIA2Driver;
+import io.appium.uiautomator2.utils.Logger;
+
+public class EnableMultiWindows extends AbstractSetting<Boolean> {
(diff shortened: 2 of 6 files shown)


app/src/main/java/io/appium/uiautomator2/model/settings/EnableMultiWindows.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    public Boolean getValue() {
+        return AppiumUIA2Driver.getInstance()
+                .getSessionOrThrow()
+                .getCapability(getName(), false);
+    }
+
+    @Override
+    protected void apply(Boolean multiWindowsEnabled) {
+        Logger.debug("Dummy setting. Maintained in Session.capabilities.");
```
mykola-mokhnach: This log is not needed

app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
```diff
@@ -85,13 +87,20 @@ public static void refreshRootAXNode() {
...
-         * TODO: MULTI_WINDOW is disabled, UIAutomatorViewer captures active window properties and
-         * end users always relay on UIAutomatorViewer while writing tests.
+         * TODO: MULTI_WINDOW is disabled by default, UIAutomatorViewer captures active window
+         * properties and end users always relay on UIAutomatorViewer while writing tests.
          * If we enable MULTI_WINDOW it effects end users.
          * https://code.google.com/p/android/issues/detail?id=207569
          */
-        if (CustomUiDevice.getInstance().getApiLevelActual() >= Build.VERSION_CODES.LOLLIPOP && MULTI_WINDOW) {
+
+        boolean multiWindowsEnabled = AppiumUIA2Driver
```
mykola-mokhnach: this should be read from the setting

app/src/main/java/io/appium/uiautomator2/model/UiAutomationElement.java
```diff
@@ -124,11 +127,7 @@ public static UiAutomationElement rebuildForNewRoot(AccessibilityNodeInfo rawEle
     }
 
     @Nullable
-    public static UiAutomationElement getCachedElement(AccessibilityNodeInfo rawElement,
-                                                       AccessibilityNodeInfo windowRoot) {
-        if (cache.get(rawElement) == null) {
-            rebuildForNewRoot(windowRoot, null);
-        }
+    public static UiAutomationElement getCachedElement(AccessibilityNodeInfo rawElement) {
```
mykola-mokhnach: In how many places this call is used? Can it break anything else if changed this way?

mykola-mokhnach: I'd. rather pass the array of window roots to keep the backward compatibility

app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
```diff
@@ -85,13 +87,20 @@ public static void refreshRootAXNode() {
...
+                .getInstance()
+                .getSessionOrThrow()
+                .getCapability(ENABLE_MULTI_WINDOWS.toString(), false);
+
+        if (CustomUiDevice.getInstance().getApiLevelActual() >= Build.VERSION_CODES.LOLLIPOP && multiWindowsEnabled) {
             // Support multi-window searches for API level 21 and up
             for (AccessibilityWindowInfo window : CustomUiDevice.getInstance().getInstrumentation()
                     .getUiAutomation().getWindows()) {
```
mykola-mokhnach: shouldn't we try to refresh the accessibility if window.getRoot == null?

app/src/main/java/io/appium/uiautomator2/model/UiAutomationElement.java
```diff
@@ -111,9 +114,9 @@ private void setDepth(int depth) {
         this.depth = depth;
     }
 
-    public static UiAutomationElement rebuildForNewRoot(AccessibilityNodeInfo rawElement, @Nullable List<CharSequence> toastMSGs) {
+    public static UiAutomationElement rebuildForNewRoot(AccessibilityNodeInfo[] rawElements, @Nullable List<CharSequence> toastMSGs) {
```
mykola-mokhnach: How about changing Root to Roots?

app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
Volune: Can you point me to an example of reading from the settings in the code? I saw that way at first

app/src/main/java/io/appium/uiautomator2/model/UiAutomationElement.java
Volune: I saw it used in only one place. When I looked at it, I didn't see the point of having that logic in UiAutomationElement. I can revert.

app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
mykola-mokhnach: https://github.com/appium/appium-uiautomator2-server/blob/f117326c6695a4dbb6382ecb6e77049df3874817/app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java#L196

app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
Volune: I guess yes. I'm not familiar with that logic, I'll have a look tomorrow

app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
```diff
@@ -85,13 +87,20 @@ public static void refreshRootAXNode() {
      */
     public static AccessibilityNodeInfo[] getWindowRoots() {
         List<AccessibilityNodeInfo> ret = new ArrayList<>();
+
         /*
-         * TODO: MULTI_WINDOW is disabled, UIAutomatorViewer captures active window properties and
-         * end users always relay on UIAutomatorViewer while writing tests.
+         * TODO: MULTI_WINDOW is disabled by default, UIAutomatorViewer captures active window
+         * properties and end users always relay on UIAutomatorViewer while writing tests.
```
mykola-mokhnach: rely on

app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
mykola-mokhnach: nevermind - clearing the accessibility cache here might invalidate previously retrieved roots. Lets keep it as is

app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
mykola-mokhnach: affected

app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
mykola-mokhnach: We could try

```java
        if (CustomUiDevice.getInstance().getApiLevelActual() >= Build.VERSION_CODES.LOLLIPOP && MULTI_WINDOW) {
            boolean wasAccessibilityRefreshed = false;
            do {
                boolean shouldRefreshAccessibility = false;
                // Support multi-window searches for API level 21 and up
                for (AccessibilityWindowInfo window : CustomUiDevice.getInstance().getInstrumentation().getUiAutomation().getWindows()) {
                    AccessibilityNodeInfo root = window.getRoot();
                    if (root == null && !wasAccessibilityRefreshed) {
                        shouldRefreshAccessibility = true;
                        break;
                    }
                    if (root == null) {
                        Logger.debug(String.format("Skipping null root node for window: %s", window.toString()));
                        continue;
                    }
                    ret.add(root);
                }

                if (!shouldRefreshAccessibility) {
                    break;
                }

                ret.clear();
                Device.waitForIdle();
                clearAccessibilityCache();
                wasAccessibilityRefreshed = true;
            } while (true);
        } else {
            ret.add(currentActiveWindowRoot());
        }
```

mykola-mokhnach: Lets keep it as is for now, we might change the implementation later if we observe any problems with it

mykola-mokhnach:
I think this change requires a but more digging than it seems.
I'll try to recreate it separately plus add the necessary changes if you don't mind

mykola-mokhnach:
I've prepared https://github.com/appium/appium-uiautomator2-server/pull/303
Can you please check if it works for you?

Volune:
I tried #303 and it was working for me (with and without option enabled). Closing this PR in favor of the other.

PR #303, opened 2019-09-11
feat: Introduce multi-window mode support

mykola-mokhnach:
The PR is based on https://github.com/appium/appium-uiautomator2-server/pull/301 plus adds some necessary refactorings


mykola-mokhnach:
Not entirely, one of the top level hierarchy branches might be similar (the one which represents the active window).

Volune:
When I test on Appium Desktop, I see the non-active windows as siblings of the active window. (The code implies that toasts are siblings at the same level, but I didn't test them).

Interestingly I can also see the system windows and keyboard. I guess later it could be useful to filter out windows that are not `TYPE_APPLICATION` and if possible add an attribute "active"

PR #323, opened 2020-01-04
fix: removed uesless files

jnhyperion:
These files are not necessary under module folder "app".


jnhyperion:
These files are not necessary under module folder "app".

mykola-mokhnach:
Thanks for the PR @jnhyperion

PR #329, opened 2020-01-29
chore: Be smarter about setting the text

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/13886


mykola-mokhnach:
yes, setText does replace the current text. This just optmizes the logic if we know the content should be replaced, so we don't perform unnecessary calls to accessibility methods.

PR #337, opened 2020-03-30
chore: Update sign-apk task to sign both server binaries

mykola-mokhnach:


KazuCocoa:
I checked this worked with build tools v28.0.3 (and install also worked)

PR #339, opened 2020-04-01
fix: Increase cross events delay to improve gestures performance

mykola-mokhnach:
I did some performance tests and it looks like the motion events generation delay of 5 ms (the same as UIAutomator framework uses) it too little for the most of devices. As the result the actual action execution period is longer than the action duration, especially on emulators. This PR increases the default value to 20ms, which improves the performance and reduces intercall delays.


mykola-mokhnach:
Kudos to @jlipps for his investigation made on this topic in https://appiumpro.com/editions/107

PR #345, opened 2020-04-17
refactor: Make gesture helpers more readable

mykola-mokhnach:


app/src/main/java/io/appium/uiautomator2/handler/Flick.java
```diff
@@ -15,13 +31,14 @@
...
 import io.appium.uiautomator2.utils.Logger;
-import io.appium.uiautomator2.utils.Point;
+import io.appium.uiautomator2.model.Point;
 import io.appium.uiautomator2.utils.PositionHelper;
 
 import static io.appium.uiautomator2.utils.Device.getUiDevice;
 import static io.appium.uiautomator2.utils.ModelUtils.toModel;
 
 public class Flick extends SafeRequestHandler {
+    private static final double SPEED_MAGIC = 1250.0;
```
KazuCocoa, line 41: It's out of curiosity, do you remember the reason why we set this magic number?
(like wait time in XCUITest)

mykola-mokhnach: no idea. This number was here from the very beginning

mykola-mokhnach: I just moved it to a constant for better readability

PR #347, opened 2020-04-20
feat: Update scrollTo handler to support custom scrollable selectors

mykola-mokhnach:


app/src/main/java/io/appium/uiautomator2/utils/Device.java
```diff
@@ -44,9 +44,11 @@ public static AndroidElement getAndroidElement(String id, Object element, boolea
         return getAndroidElement(id, element, isSingleMatch, null, null);
     }
 
-    public static void scrollToElement(UiSelector selector, int maxSwipes)
-            throws UiObjectNotFoundException {
-        UiScrollable uiScrollable = new UiScrollable(new UiSelector().scrollable(true).instance(0));
+    public static void scrollToElement(@Nullable UiScrollable origin, UiSelector selector,
+                                       @Nullable Integer maxSwipes) throws UiObjectNotFoundException {
+        UiScrollable uiScrollable = origin == null
+                ? new UiScrollable(new UiSelector().scrollable(true).instance(0))
```
KazuCocoa, line 50: `new UiSelector().scrollable(true).instance(0)`'s instance number can be greater than 1.
What about leaving a class name and resource id as a log which element uia2 server will scroll?

mykola-mokhnach: np, will add

KazuCocoa:
appium-uiautomator2-server@4.8.0

PR #350, opened 2020-05-05
Fix getText bug for UiObject element

AlexandruMaxim:
For an UiObject element (e.g. Seekbar) the text value returned is not correct.
The page source contains the correct value but for a request of type getText,
the value is different.


app/src/main/java/io/appium/uiautomator2/utils/ElementHelpers.java
```diff
@@ -198,9 +198,9 @@ public static String getText(Object element, boolean replaceNull) throws UiObjec
                 return charSequenceToString(nodeInfo.getText(), replaceNull);
             }
```
mykola-mokhnach, line 200: extra empty line here

AlexandruMaxim: Solved, thank you

Issue #358, opened 2020-06-10
25% slowdown in uiautomator2 server version 2.6.0


sshock:
We've been using Appium 1.9.1 (uiautomator2 1.18.0) for almost 2 years now (bad! we need to update!).

Unfortunately, something changed after Appium 1.9.1 that caused most of our automation jobs to take about 25% longer (which is a big deal for us).  I tested every version of Appium up to the latest (1.17.1) and it never goes back down to the speed of 1.9.1.

Here is a table showing the slowdown after Appium 1.9.1 that can be seen in almost every automation job type except Job A.  (Each job was ran three times; the times are in seconds).

Released | Appium Version | Job A | Job B | Job C | Job D | Job E
-- | -- | -- | -- | -- | -- | --
Sep 2018 | 1.9.1 | 27,30,30 | 43,43,47 | 43,44,43 | 49,44,46 | 47,45,49
Nov 2018 | 1.10.0 | (install failed) ||||
Feb 2019 | 1.11.0 | 33,32,32 | 97,82,110 | 63,61,64 | 73,63,74 | 56,62,56
Mar 2019 | 1.12.0 | 63,35,39 | 93,54,60 | 89,52,? | 88,63,65 | 75,59,57
May 2019 | 1.13.0 | 31,30,31 | 66,56,60 | 39,52,46 | 59,73,61 | 59,65,64
Aug 2019 | 1.14.2 | 37,30,38 | 64,58,67 | 51,48,43 | 62,59,58 | 57,59,58
Oct 2019 | 1.15.1 | 30,29,31 | 78,63,79 | 46,46,45 | 57,59,55 | 61,54,56
Dec 2019 | 1.16.0 | 72,29,29 | 103,45,64 | 75,45,48 | 89,59,53 | 85,61,56
Mar 2020 | 1.17.0 | 30,31,28 | 71,71,72 | 48,46,? | 72,58,54 | 52,56,57
May 2020 | 1.17.1 | 32,28,32 | 68,61,61 | n/a | 67,62,57 | 58,58,52

So I decided to test uiautomator2-server from version 1.18.0 to 3.1.1 (the versions that correspond to Appium 1.9.1 to 1.11.0).  I tested with Job B, but using a faster device.  This test indicates that the slowdown occurred in version 2.6.0 of appium-uiautomator2-server.

version | status
-- | --
1.18.0 | GOOD (30s)
1.19.0 | GOOD (29s)
1.19.1 | GOOD (30s)
[...]

PR #360, opened 2020-06-13
feat: Only include the actually queried attributes into the source xml

mykola-mokhnach:
https://github.com/appium/appium-uiautomator2-server/issues/358#issuecomment-643531422

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         maven { url 'https://jitpack.io' }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.0.0-rc01'
+        classpath 'com.android.tools.build:gradle:4.0.0'
         classpath 'com.github.JakeWharton:sdk-manager-plugin:0ce4cdf08009d79223850a59959d9d6e774d0f77'
         classpath 'de.mobilej.unmock:UnMockPlugin:0.6.4'
     }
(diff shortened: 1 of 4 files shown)


sshock:
Will test on Monday...

sshock:
Initial testing looks pretty good.

time (seconds) | uiautomator2 version
-- | --
56, 75, 55, 66 | * version 4.5.5
56, 56, 57, 55 | * master (version 4.8.3)
52, 50, 54 | * xml_attr branch (this MR)

I will plan on doing some more testing with more job types tomorrow.

sshock:
Here are some more test results with various types of jobs we do.

uiaut2 version | Job A | Job B | Job C | Job D | Job E
-- | -- | -- | -- | -- | --
4.5.5 | 32,28,32 | 68,61,61 | 64,50,55 | 67,62,57 | 58,58,52
4.8.3 | 32,29,29 | 80,68,59 | 52,53,52 | 58,59,64 | 59,59,58
xml_attr branch | 31,31,27 | 58,52,56 | 57,51,52 | 66,58,57 | 68,51,51

I think it is safe to say this change does help a bit, with Job type B at least.  So it seems like a good change to me that can be merged.

The bad news is, this change alone doesn't seem to fully solve #358.  All of the job types (including Job B) are still noticeably slower than they were in Appium 1.9.1 (uiautomator2-server 1.18.0), as seen in the table at the top of that issue.

mykola-mokhnach:
Thanks for checking this. Many things were changed since that version. Also, I would say it makes sense to sacrifice speed to correctness. @sshock feel free to share your further observations on that. We would be happy to help with what we can.

PR #391, opened 2020-11-08
add Page Views

xdvrx1:
When we add this service, it will count every hit of this repo. And this will guide us more about the visitors each day and will indicate the total views. For me, this is very helpful both for us and for those will view this repo seeing this page views. If there is the website built from this repo, it can be simply added there too.

Screenshot (1705)


xdvrx1:
I'll read the agreement. I may or may not sign.

KazuCocoa:
Thanks!
We'll wait for the CLA

mykola-mokhnach (changes requested):
I don't like the idea of having additional tracking (and don't like the idea of tracking in general). Currently it is fine to have the statistics that GitHub itself provides

xdvrx1:
Yup, that was the answer when I was not satisfied by the GitHub Traffic. This one is a simple counter for every page hit. The most important one is, it's displayed on your page or repo README file, so that others will know also the visitors per day and the total count. It does not fetch data from either Google Analytics or GitHub Traffic.

You can check out the service too if there is doubt,

<https://hits.seeyoufarm.com/>

the development is in GitHub too:

<https://github.com/gjbae1212/hit-counter>

mykola-mokhnach:
Sorry, but we have decided to not merge such PRs

Issue #405, opened 2021-01-18
Finding elements is considerably slower after version 4.16.0


jatalahd:
We have used Appium v1.17.1 in Android app testing until now and we have been very happy with it. Now we started to evaluate upgrading to the latest Appium version v1.20.1. Unfortunately we noticed regression in element location speed, which in our case adds 2 hours of execution time to our whole test suite.

I narrowed the problem to this component by separately installing different versions of appium-uiautomator2-server while keeping Appium version at 1.20.1. The last appium-uiautomator2-server version, where the elements are found in decent time was version 4.16.0. After that version the timing issues become evident in our tests. The problem arises most clearly in situations, where all elements from a list view are read using find_elements_by_id(locator). The timing regression in these cases can be measured in several seconds, even up to 10 seconds per one call to find_elements() in our case. Also it seems that the more calls there is to find_elements(), it builds up slower performance as time goes on.

I see that there has been done some refactoring to element locator handling after version 4.16.0 was released. I would hope that the performance of finding elements would be returned to the same level as it was before. Otherwise we cannot use the latest versions of Appium anymore.

mykola-mokhnach:
Thanks for the report @jatalahd 
Please try if the patch https://github.com/appium/appium-uiautomator2-server/pull/406 works for you

jatalahd:
Thank you for the quick fix!

I Tested most of our test suites with patch #406 built on top of the master branch (version 4.17.3). Test execution times returned back to normal level and no timing issues observed in tests.

Verification Summary: Fix OK from my point of view.

mykola-mokhnach:
Cool, thanks for your help. Unfortunately this is one of the issues, which are quite hard to detect and avoid.

mykola-mokhnach:
The patch has been published to appium@beta

PR #406, opened 2021-01-18
fix: Improve caching performance

mykola-mokhnach:
It looks like calling `equals` on UiObject is pretty expensive, since it tries to find the connected accessibility element. Related to https://github.com/appium/appium-uiautomator2-server/issues/405


KazuCocoa (approved):

tommeier:
Thank you so much for fixing this 🙏 I've been hunting this issue for the last couple of days and this has brought our builds from 45 mins back down to normal 25min range for android ❤️

PR #410, opened 2021-01-26
fix: Update the way fetchElements handles INSTANCE selector

robertstreche:
Changed the fetchElements() method to check if the selector "contains" "INSTANCE" instead of "ends with" "INSTANCE" attribute.

An example of the wrong behavior can be observed using the following selector:
UiSelector[CLASS=android.widget.CheckedTextView, INSTANCE=2, RESOURCE_ID=android:id/text1]

The current version of fetchElements() method checks if the selector ends with "INSTANCE", in which case the instance attribute will be ignored.

In order to solve this problem the method could check if the selector contains "INSTANCE" and the instance attribute will be used into the element selection.

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/handler/FindElements.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/FindElements.java
@@ -53,7 +53,7 @@ import static io.appium.uiautomator2.utils.StringHelpers.isBlank;
 
 public class FindElements extends SafeRequestHandler {
 
-    private static final Pattern endsWithInstancePattern = Pattern.compile(".*INSTANCE=\\d+]$");
+    private static final Pattern containsInstancePattern = Pattern.compile(".*INSTANCE=\\d+.*");
 
     public FindElements(String mappedUri) {
         super(mappedUri);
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/appium/uiautomator2/handler/FindElements.java
```diff
@@ -182,7 +182,7 @@ protected AppiumResponse safeHandle(IHttpRequest request) throws UiObjectNotFoun
         // UiSelector[CLASS=android.widget.Button, INSTANCE=1]
         //
         // The selector now points to an entirely different element.
-        if (endsWithInstance) {
+        if (containsInstance) {
             Logger.debug("Selector ends with instance.");
```
mykola-mokhnach, line 186: Please update the log string and the comment above

app/src/main/java/io/appium/uiautomator2/handler/FindElements.java
```diff
@@ -170,7 +170,7 @@ protected AppiumResponse safeHandle(IHttpRequest request) throws UiObjectNotFoun
         boolean keepSearching = true;
```
mykola-mokhnach, line 170: It would be nice to extract this helper into a separate class and cover it with unit tests. Ofc not in this PR

Do you have time to help with that @robertstreche ?

KazuCocoa:
Could you sign the cla as well?

mykola-mokhnach:
@robertstreche I have prepared a bigger refactoring, which also includes this change in https://github.com/appium/appium-uiautomator2-server/pull/411

Could you please help us to test that change properly and confirm that it works for all your test scenarious before we merge it? Thanks

app/src/main/java/io/appium/uiautomator2/handler/FindElements.java
robertstreche: The PR [https://github.com/appium/appium-uiautomator2-server/pull/411](url) seems to be just the refactoring you are talking about.

mykola-mokhnach: yes, I wanted to do such refactoring long time ago, just did not have a good test suite to verify it on.

mykola-mokhnach:
Closed as subset of https://github.com/appium/appium-uiautomator2-server/pull/411

PR #411, opened 2021-01-26
refactor: Update ByUiAutomator lookup approach

mykola-mokhnach:
Related to https://github.com/appium/appium-uiautomator2-server/pull/410

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/common/exceptions/UiSelectorSyntaxException.java
+++ b/app/src/main/java/io/appium/uiautomator2/common/exceptions/UiSelectorSyntaxException.java
@@ -19,22 +19,26 @@ package io.appium.uiautomator2.common.exceptions;
 import androidx.annotation.Nullable;
 
 public class UiSelectorSyntaxException extends InvalidSelectorException {
-
-    private static String formatErrorMessage(String expression, String message) {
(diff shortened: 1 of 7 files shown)


KazuCocoa (approved):
seems good to me.
(Haven't tested with checking out this code for now, but CI was passed.. 🍏 )

robertstreche:
Reffering to [https://github.com/appium/appium-uiautomator2-server/pull/410](url)

I've tested using your PR and the following selector:
`UiSelector[CLASS=android.widget.CheckedTextView, INSTANCE=2, RESOURCE_ID=android:id/text1]`

I have the `java.lang.NoSuchFieldException: SELECTOR_INSTANCE` exception when searching for the INSTANCE attribute.

According to the java documentation, it seems to be a problem into this line of code: 
`Field field = UiSelector.class.getField(fieldName);`
The problem is that getField(String) method returns NoSuchFieldException SELECTOR_INSTANCE and it seems to be a problem related to 'default static final' access modifiers of the SELECTOR_* fields.

Using getDeclaredField(String) seems to work fine, but I cannot guarantee it is the right way to resolve the problem.,

mykola-mokhnach:
Thanks for the hint @robertstreche 
I've added the necessary corrections.

PR #420, opened 2021-04-14
chore: Simplify retrieval of system metrics

mykola-mokhnach:
Could also resolve https://github.com/appium/appium/issues/15154


KazuCocoa (approved):
Lg.
Could this improve the instrumentation process crash as you addressed?

mykola-mokhnach:
Not guaranteed, but it won't make the thing worse for sure

PR #425, opened 2021-05-15
chore: Add better safety against NullPointerException

mykola-mokhnach:


app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
```diff
@@ -91,27 +90,14 @@ public int getApiLevelActual() {
         return (Integer) API_LEVEL_ACTUAL;
     }
 
-    @Nullable
     private UiObject2 toUiObject2(Object selector, AccessibilityNodeInfo node) {
         Object[] constructorParams = {getUiDevice(), selector, node};
-        long end = SystemClock.uptimeMillis() + UIOBJECT2_CREATION_TIMEOUT;
-        while (true) {
```
KazuCocoa, line 98: Don't need to wait for the element?

KazuCocoa (approved):
lg

app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
mykola-mokhnach: nope, this constructor is never going to throw an exception. And retry does not help anyway even if it does

PR #427, opened 2021-05-19
chore: skip idle checks while retrieving cached node from uiobject2

mykola-mokhnach:
The `getAccessibilityNodeInfo` method of UiObject2 contains calls for `getDevice().waitForIdle()` and `mCachedNode.refresh()`, which are redundant for the case when we want to extract the cached node from a just created UiObject2 instance (and also may affect performance in some cases).

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/core/AxNodeInfoExtractor.java
+++ b/app/src/main/java/io/appium/uiautomator2/core/AxNodeInfoExtractor.java
@@ -26,6 +26,7 @@ import androidx.test.uiautomator.UiObject2;
 
 import io.appium.uiautomator2.common.exceptions.StaleElementReferenceException;
 
+import static io.appium.uiautomator2.utils.ReflectionUtils.getField;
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/appium/uiautomator2/core/AxNodeInfoExtractor.java
```diff
@@ -36,6 +37,13 @@ public static AccessibilityNodeInfo toNullableAxNodeInfo(Object object) {
         return extractAxNodeInfo(object);
     }
 
+    @Nullable
+    public static AccessibilityNodeInfo toNullableAxNodeInfo(UiObject2 object, boolean checkStaleness) {
+        return checkStaleness
+                ? extractAxNodeInfo(object)
+                : (AccessibilityNodeInfo) getField("mCachedNode", object);
```
KazuCocoa, line 44: 👍

KazuCocoa (approved):
Read the code of UiObject2. Actually `mCachedNode` seems a good idea

PR #432, opened 2021-07-14
Changing uiautomator to use decoration-less screen info

Kertic:
https://developer.android.com/reference/android/support/test/uiautomator/UiDevice#getDisplaySizeDp()

Changing the result to use the display size independent of decorations to allow tapping on elements in fullscreen contexts that would normally be covered by decorations


Kertic:
https://github.com/appium/appium/issues/11244
Looking to solve this issue. It's been resting and automatically closing when it comes up, I'd like this to work.

Kertic:
Looking at the unit tests I'm seeing a failure on the device commands test, but I'm unable to identify why it failed as I can't access any artifacts from the azure portal

mykola-mokhnach:
This is going to be a breaking change, we cannot accept it. Consider rather changing APIs where this value is checked

Kertic:
> This is going to be a breaking change, we cannot accept it. Consider rather changing APIs where this value is checked

This makes sense, the unit tests broke. Will look into other methods.

PR #435, opened 2021-07-15
Changing functionality to reflect decorationless sizes

Kertic:
Testing these changes with the github PR automated testing, the goal of this is to allow the device to gather the device bounds in contexts where the device bounds are desired, instead of including decorations

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/utils/PositionHelper.java
+++ b/app/src/main/java/io/appium/uiautomator2/utils/PositionHelper.java
@@ -68,7 +68,8 @@ public abstract class PositionHelper {
 
     public static Point getDeviceAbsPos(final Point point) {
         final UiDevice d = UiDevice.getInstance(getInstrumentation());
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
Could you please sign CLA?

Also have you tested this change locally? Does the click happen where it is supposed to happen?

Kertic:
image

I'm unable to sign it in its current state due to the attached image, and I was not able to verify on an actual device yet. We can hold off on merging pending proven local testing, I'll just need to figure out how to build and align everything locally. Are there resources that can guide me in setting that up?

mykola-mokhnach:
The README.md contains basic info on how to build the server. Then you could simply put the compiled apk files into an existing Appium deployment and that will work.

Kertic:
I will do that when time permits and report back if it succeeds or comment and close the PR if it fails. As for signing the CLA, I can't do it directly through the "Details" link due to the attached image, so is there another way I can agree to it that would let the system detect it?

mykola-mokhnach:
Closed in favour of https://github.com/appium/appium-uiautomator2-server/pull/441

PR #440, opened 2021-09-13
refactor: Simplify API calls where the timeout is always zero

mykola-mokhnach:


app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
```diff
@@ -119,7 +119,7 @@ public AccessibleUiObject findObject(Object selector) throws UiAutomator2Excepti
             node = (AccessibilityNodeInfo) selector;
             selector = toSelector(node);
         } else if (selector instanceof UiSelector) {
-            return toAccessibleUiObject(getUiDevice().findObject((UiSelector) selector), null);
```
KazuCocoa, line 122: Can be removed...?

mykola-mokhnach: yes. Basically providing null means we want to wait for the downstream method up to `Configurator.getInstance().getWaitForSelectorTimeout()` seconds if the given selector does not exist. This basically repeats the purpose of an implicit wait, which we have in the upstream driver. E.g. keeping null here would add unnecessary delays while looking for UiSelector's, which don't exist

KazuCocoa (approved):

app/src/main/java/io/appium/uiautomator2/core/AxNodeInfoExtractor.java
```diff
@@ -32,47 +31,35 @@
 
 public abstract class AxNodeInfoExtractor {
 
-    // The timeout argument only affects the lookup timeout for UiObject instances
-    // null timeout value means to use the default
-    // `Configurator.getInstance().getWaitForSelectorTimeout()` timeout which equals to 10000ms
-
     @Nullable
-    public static AccessibilityNodeInfo toNullableAxNodeInfo(Object object, @Nullable Long timeoutMs) {
```
KazuCocoa, line 40: oh, i see

mykola-mokhnach:
@KazuCocoa Could you please add these recent patches to the next rc?

KazuCocoa:
ok, let me update shortly

KazuCocoa:
https://www.npmjs.com/package/appium/v/1.22.0-rc.5

PR #452, opened 2022-01-08
feat: Add a possibility to enforce Xpath1 usage

mykola-mokhnach:
Unfortunately we don't get any responses from Eclipse maintainers regarding Psychopath xpath2 parser issues, so it makes sense to have a possibility to fallback to xpath1 if we face a bug for which no workaround exists.
The PR adds a new `enforceXPath1` setting, which being set to true tells the server to use XPath1 parser rather than XPath2 for xpath locator types.
Related to https://github.com/appium/appium/issues/16142#issuecomment-1003954166


KazuCocoa (approved):
yea... hope most of clients can cover their tests with XPath2

PR #458, opened 2022-01-28
Allow finding parent elements

Karazum:


mykola-mokhnach:
Unfortunately we cannot add this change as it would break the backward compatibility. Also any other selenium drivers allow that.

mykola-mokhnach (changes requested):
.

Karazum:
@mykola-mokhnach I know it looks dirty, but it is a needed feature. 
I would like to address all backward compatibility issues, so please raise your concerns.


Also as for other selenium drivers, they do allow it (but not for the xpath expression I have put in the code I can change that)
The left side is a android appium session and the right side is a selenium session.
As you can see in the screenshot the you can access parent elements while in appium you cant.
Screenshot from 2022-01-28 23-47-00

mykola-mokhnach:
I tried to implement this feature in a more flexible way: https://github.com/appium/appium-uiautomator2-server/pull/459/

Please test it locally and let me know if such approach works

mykola-mokhnach:
Appium 1 is EOL. It is only possible to get the recent drivers working with Appium 2

mykola-mokhnach:
Just to confirm: did you test it according to the the recent [comment](https://github.com/appium/appium-uiautomator2-server/pull/459#issuecomment-1024914217) (e.g. set the corresponding setting to `false`)?

Karazum:
> Just to confirm: did you test it according to the the recent [comment](https://github.com/appium/appium-uiautomator2-server/pull/459#issuecomment-1024914217) (e.g. set the `limitXPathContextScope` setting to `false`)?

I did not. I will do another round of testing on monday with appium 2.

PR #459, opened 2022-01-29
feat: Add a possibility to access parent elements in context-based searches

mykola-mokhnach:
The PR reassembles the idea from https://github.com/appium/appium-uiautomator2-server/pull/458

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/core/AccessibilityNodeInfoDumper.java
+++ b/app/src/main/java/io/appium/uiautomator2/core/AccessibilityNodeInfoDumper.java
@@ -24,6 +24,7 @@ import android.view.Display;
 import android.view.View;
 import android.view.accessibility.AccessibilityNodeInfo;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import org.apache.commons.io.IOUtils;
@@ -99,7 +100,7 @@ public class AccessibilityNodeInfoDumper {
         this.includedAttributes = includedAttributes;
     }
 
-    private String buildRootElementSearchQuery() {
+    private int matchRootElementIndex() {
         if (root == null) {
             throw new RuntimeException(
                     "Root node search query cannot be built if the root is unset"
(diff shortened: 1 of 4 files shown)


mykola-mokhnach:
It looks like adding the feature as is would break the backward compatibility, because Xpath parser ignores the current context if a query starts with `//`, so existing queries that look like `//element` might start returning different results if executed within a context. That is why a new setting called `limitXPathContextScope` has been added. This setting is set to `true` by default and resembles the legacy behavior. Setting it to false makes it possible to execute elements location through the whole page source, although one has to keep in mind, that the search root must be always provided for queries to work properly (e.g. queries like `//element` must be replaced with `.//element`).

Karazum:
Unfortunately, I did not manage to test it with appium 2.0

I have tested this new settings, it is working as expected. If set to true, only child elements can be found, therefore preserving the backward compatibility.

With the flag set to false, parent elements can be retrieved and child elements can be queried with xpath that starts with `./`

mykola-mokhnach:
Thanks @Karazum 
So you confirm the current implementation fixes your issue and aligns uia2 driver behaviour with other selenium drivers?

Karazum:
> Thanks @Karazum So you confirm the current implementation fixes your issue and aligns uia2 driver behaviour with other selenium drivers?

Yeah, thank you very much. :)

On a semi related note, I think the ios [web driver agent](https://github.com/appium/WebDriverAgent) also has this issue.

mykola-mokhnach:
I think all Appium drivers have it. As you can see the fix is not very straightforward and requires good knowledge of the used frameworks. Nevertheless feel free to implement it for WDA. I'd be happy to review

Karazum:
> I think all Appium drivers have it. As you can see the fix is not very straightforward and requires good knowledge of the used frameworks. Nevertheless feel free to implement it for WDA. I'd be happy to review

I will give it a try in a week, but my knowledge of Objective C and iOS drivers is very poor.

app/src/main/java/io/appium/uiautomator2/core/AccessibilityNodeInfoDumper.java
```diff
@@ -98,6 +101,52 @@ public AccessibilityNodeInfoDumper(@Nullable AccessibilityNodeInfo root,
...
+        }
+        return null;
+    }
+
+    @NonNull
+    private Node fetchContext(InputStream xml) {
+        Document doc = loadDocument(xml);
+        return root == null || Settings.get(LimitXpathContextScope.class).getValue()
+                ? doc
+                : Objects.requireNonNull(
```
KazuCocoa, line 144: okay.. so in this case current element can be considered as https://github.com/appium/appium-uiautomator2-server/pull/459#issuecomment-1024914217
btw, do we have test code related to this change?
Maybe in uia2-driver's e2e or this repo's test code may help our future

KazuCocoa (approved):

app/src/main/java/io/appium/uiautomator2/core/AccessibilityNodeInfoDumper.java
mykola-mokhnach: this test would be a bit more complicated as we need to switch settings to check it. Currently e2e tests are disabled completely for all drivers since they need to be rewritten to comply with W3C/WDIO

PR #460, opened 2022-02-23
feat: add hint attribute for API Level 26+

KazuCocoa:
not tested yet. Will do later.

https://developer.android.com/reference/android/view/accessibility/AccessibilityNodeInfo#getHintText() is available since api 26.

https://github.com/appium/appium/discussions/16484

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/model/UiObjectElement.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/UiObjectElement.java
@@ -129,6 +129,8 @@ public class UiObjectElement extends BaseElement {
             case HINT:
                 if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
                     result = toAxNodeInfo(element).getHintText();
+                } else {
+                    result = null;
                 }
                 break;
             case SELECTION_END:
--- a/app/src/main/java/io/appium/uiautomator2/model/UiObjectElement.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/UiObjectElement.java
@@ -16,6 +16,7 @@
 
 package io.appium.uiautomator2.model;
 
+import android.os.Build;
 import android.util.Pair;
 import android.view.accessibility.AccessibilityNodeInfo;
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/io/appium/uiautomator2/model/UiObjectElement.java
```diff
@@ -125,6 +126,13 @@ public String getAttribute(String attr) throws UiObjectNotFoundException {
                 result = AxNodeInfoHelper.getPackageName(toAxNodeInfo(element));
                 break;
             }
+            case HINT:
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
mykola-mokhnach, line 130: could it use a ternary operator?

app/src/main/java/io/appium/uiautomator2/utils/Attribute.java
```diff
@@ -41,6 +41,7 @@
     SELECTION_END(new String[]{"selection-end"}),
     SELECTED(new String[]{"selected"}),
     TEXT(new String[]{"text", "name"}),
+    HINT(new String[]{"text", "hint"}),
```
mykola-mokhnach, line 44: Please remove "text" from this array as it might create a serious confusion in the downstream code if other items have the same alias

app/src/main/java/io/appium/uiautomator2/model/UiObjectElement.java
KazuCocoa: I thought so, but AS recommended this style, so just followed it to reduce warning by IDE at this time

app/src/main/java/io/appium/uiautomator2/utils/Attribute.java
KazuCocoa: yea, that was my wrong

mykola-mokhnach (approved):

KazuCocoa:
I found the `hint` in a google map app. XPath also worked.
```
[6] pry(#<AppiumLibCoreTest::WebDriver::DeviceTest>)> e= @driver.find_element :xpath, '//*[@hint="ここで検索"]'
=> #<Appium::Core::Element:0x2d330cf992039026 id="00000000-0000-002a-ffff-ffff00000043">
[7] pry(#<AppiumLibCoreTest::WebDriver::DeviceTest>)> e.hint
=> "ここで検索"
```

Issue #462, opened 2022-03-29
enableMultiWindows and disableSuppressAccessibilityService not working simultaneously


nurdtechie98:
When trying to run a session, with`enableMultiWindows` and `disableSuppressAccessibilityService` set to true, none of the window related commands including find element work. 
Find element returns a 500 response 
```
2022-03-25 12:56:58:810 - [HTTP] --> POST /wd/hub/session/635dbc22-ebf9-4bcd-bc4b-18def21d2cbe/element
2022-03-25 12:56:58:810 - [HTTP] {"using":"id","value":"connectivity_status"}
2022-03-25 12:56:58:811 - [debug] [MJSONWP (635dbc22)] Calling AppiumDriver.findElement() with args: ["id","connectivity_status","635dbc22-ebf9-4bcd-bc4b-18def21d2cbe"]
2022-03-25 12:56:58:811 - [debug] [BaseDriver] Valid locator strategies for this request: xpath, id, class name, accessibility id, css selector, -android uiautomator
2022-03-25 12:56:58:811 - [debug] [BaseDriver] Waiting up to 0 ms for condition
2022-03-25 12:56:58:812 - [debug] [WD Proxy] Matched '/element' to command name 'findElement'
2022-03-25 12:56:58:813 - [debug] [WD Proxy] Proxying [POST /element] to [POST http://127.0.0.1:8203/wd/hub/session/51936b19-2176-4aca-ac9b-d265afa3a53f/element] with body: {"strategy":"id","selector":"connectivity_status","context":"","multiple":false}
[...]

PR #463, opened 2022-03-30
fix: Pass Uiautomator flags to all instances

nurdtechie98:
Related to https://github.com/appium/appium-uiautomator2-server/issues/462

As a part of https://github.com/appium/appium-uiautomator2-server/pull/289 appium started supporting [FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES](https://developer.android.com/reference/android/app/UiAutomation#FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES)

In order to do so the Configurator instance is initialised here 
https://github.com/appium/appium-uiautomator2-server/blob/d3cd4a3baa4209d02afcdd2fc0007717890bf138/app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java#L69-L70

While this configurator comes into affect whenever uiautomation instance is used from UiAutomatorBridge
https://github.com/appium/appium-uiautomator2-server/blob/700eefebb75f8c7e2d3d1feac3b6d0d13e8b339d/app/src/main/java/io/appium/uiautomator2/core/UiAutomatorBridge.java#L52

But there are several instances where fresh instances of uiautomator are initialized using `CustomUiDevice` like
https://github.com/appium/appium-uiautomator2-server/blob/700eefebb75f8c7e2d3d1feac3b6d0d13e8b339d/app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java#L88-L91

Since in this cases a new instance of uiautomator is created directly from `Instrumentation`. The instance the doesn't read the flags from configurator and initializes with default flags, which would be `FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES` as false thus stopping all accessibility services created and starts behaving in a weird manner.

As a part of this PR, creating a custom `getUiAutomation` method in `CustomUiDevice` which before instantiating the uiautomator checks the configurator flags, and use that method at all the places were it was previously being used.

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
@@ -90,8 +90,8 @@ public class CustomUiDevice {
     }
 
     public UiAutomation getUiAutomation() {
-        int flags = Configurator.getInstance().getUiAutomationFlags();
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            int flags = Configurator.getInstance().getUiAutomationFlags();
             return getInstrumentation().getUiAutomation(flags);
         } else {
             return getInstrumentation().getUiAutomation();
--- a/app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
@@ -93,9 +93,8 @@ public class CustomUiDevice {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
             int flags = Configurator.getInstance().getUiAutomationFlags();
             return getInstrumentation().getUiAutomation(flags);
-        } else {
-            return getInstrumentation().getUiAutomation();
         }
+        return getInstrumentation().getUiAutomation();
     }
 
     public int getApiLevelActual() {
--- a/app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
@@ -17,6 +17,8 @@
 package io.appium.uiautomator2.model.internal;
 
 import android.app.Instrumentation;
+import android.app.UiAutomation;
+import android.os.Build;
 import android.os.SystemClock;
 import android.view.accessibility.AccessibilityNodeInfo;
 
@@ -26,6 +28,7 @@ import androidx.test.uiautomator.BySelector;
 import androidx.test.uiautomator.UiDevice;
 import androidx.test.uiautomator.UiObject2;
 import androidx.test.uiautomator.UiSelector;
+import androidx.test.uiautomator.Configurator;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
@@ -84,6 +87,16 @@ public class CustomUiDevice {
 
     public Instrumentation getInstrumentation() {
         return mInstrumentation;
+    }
+
+    public UiAutomation getUiAutomation() {
+        int flags = Configurator.getInstance().getUiAutomationFlags();
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N)
+        {
+            return getInstrumentation().getUiAutomation(flags);
+        } else {
+            return getInstrumentation().getUiAutomation();
+        }
     }
 
     public int getApiLevelActual() {
@@ -173,7 +186,7 @@ public class CustomUiDevice {
             return desired;
         }
 
-        getInstrumentation().getUiAutomation().setRotation(desired.ordinal());
+        getUiAutomation().setRotation(desired.ordinal());
         long start = System.currentTimeMillis();
         do {
             if (ScreenRotation.current() == desired) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
```diff
@@ -86,6 +89,16 @@ public Instrumentation getInstrumentation() {
         return mInstrumentation;
     }
 
+    public UiAutomation getUiAutomation() {
+        int flags = Configurator.getInstance().getUiAutomationFlags();
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N)
+        {
```
mykola-mokhnach, line 95: Please format this block, so the opening bracket is on the same line with `if` operator

app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
```diff
@@ -86,6 +89,16 @@ public Instrumentation getInstrumentation() {
...
+            return getInstrumentation().getUiAutomation(flags);
```
mykola-mokhnach, line 95: I assume we only need to fetch flags inside of this block

app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
nurdtechie98: Thanks for the review, fixed this in the latest commit https://github.com/appium/appium-uiautomator2-server/pull/463/commits/162cf19fa825233529073fa04f1d770d1aaffb05

app/src/main/java/io/appium/uiautomator2/server/mjpeg/MjpegScreenshotStream.java
```diff
@@ -119,8 +118,15 @@ private void matchFramerate(long targetInterval, long start) {
         }
     }
 
+    private static synchronized UiAutomation getUiAutomation() {
+        if (UI_AUTOMATION == null) {
```
mykola-mokhnach, line 122: caching must not be applied there. Please change this method body to `return CustomUiDevice.getInstance().getUiAutomation()`

app/src/main/java/io/appium/uiautomator2/server/mjpeg/MjpegScreenshotStream.java
```diff
@@ -40,7 +40,7 @@
     private static final int NO_CLIENTS_CONNECTED_SLEEP_TIME_MS = 500;
     private static final byte[] END = "\r\n\r\n".getBytes(UTF_8);
     private static final UiAutomation UI_AUTOMATION =
-        CustomUiDevice.getInstance().getInstrumentation().getUiAutomation();
+        CustomUiDevice.getInstance().getUiAutomation();
```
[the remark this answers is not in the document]
nurdtechie98: Right, moved it a method to initialize in runtime https://github.com/appium/appium-uiautomator2-server/pull/463/commits/7fa2e1596b5bfc44c3c6ce63c4a017a4a8e846dd

Can you check once ?

app/src/main/java/io/appium/uiautomator2/server/mjpeg/MjpegScreenshotStream.java
nurdtechie98: Done, removed the method in https://github.com/appium/appium-uiautomator2-server/pull/463/commits/c2be84093da4ea9e8c9138c277664d000f60574f

KazuCocoa (approved):
thanks for your investigation! It looks good to me.

app/src/main/java/io/appium/uiautomator2/utils/ScreenshotHelper.java
```diff
@@ -48,7 +48,7 @@
 public class ScreenshotHelper {
     private static final int PNG_MAGIC_LENGTH = 8;
     private static final UiAutomation uia =
```
mykola-mokhnach, line 50: the same should be done to the static member here. UiAutomation instance must be retrieved each time upon request

app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
```diff
@@ -86,6 +89,15 @@ public Instrumentation getInstrumentation() {
         return mInstrumentation;
     }
 
+    public UiAutomation getUiAutomation() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            int flags = Configurator.getInstance().getUiAutomationFlags();
+            return getInstrumentation().getUiAutomation(flags);
+        } else {
```
mykola-mokhnach, line 96: this else is redundant

app/src/main/java/io/appium/uiautomator2/utils/ScreenshotHelper.java
nurdtechie98: Missed this, updated in the latest commit https://github.com/appium/appium-uiautomator2-server/pull/463/commits/4be80049b1d5c8ebbe2959b4068273eccef725a1

app/src/main/java/io/appium/uiautomator2/model/internal/CustomUiDevice.java
nurdtechie98: removed else https://github.com/appium/appium-uiautomator2-server/pull/463/commits/4be80049b1d5c8ebbe2959b4068273eccef725a1

mykola-mokhnach (approved):

nurdtechie98:
Thanks, @mykola-mokhnach and @KazuCocoa for thorough and quick reviews 🙌

PR #471, opened 2022-05-17
Remove restriction of PointerDown before PointerUp

jatalahd:
- To enable long press functionality without any
  preset timeouts, it is now possible to trigger PointerDown
  in one ActionChain, do stuff in between and release the
  button using PointerUp in other ActionChain.

- Example usage with python:

  from appium import webdriver
  from selenium.webdriver.common.action_chains import ActionChains

  ActionChains(driver1).click_and_hold(ptt_button).perform()
  print ("Free to do stuff here!!")
  ActionChains(driver1).release(ptt_button).perform()

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/utils/w3c/ActionsTokenizer.java
+++ b/app/src/main/java/io/appium/uiautomator2/utils/w3c/ActionsTokenizer.java
@@ -403,11 +403,8 @@ public class ActionsTokenizer {
                 }
                 break;
                 case ACTION_ITEM_TYPE_POINTER_UP: {
-                    if (!isPointerDown) {
-                        throw new ActionsParseException(String.format(
-                                "You cannot perform '%s' action without performing '%s' first at " +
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
Have you tested this change with your automation script? The exception won't be raised, although is the touch action itself executed as expected?

Also please sign the CLA, so we could merge the PR

jatalahd:
Yes, I compiled the code, replaced the apks to the correct directory and run my test successfully. So the modification is tested. But the final obstacle seems to be that CLA. Not very Easy... I have signed it now 3 times, but nothing changes here. My computer is behind corporate proxy and cannot access the signature server from here. When signing from my mobile, it does not seem to have any effect.

mykola-mokhnach:
> Yes, I compiled the code, replaced the apks to the correct directory and run my test successfully. So the modification is tested. But the final obstacle seems to be that CLA. Not very Easy... I have signed it now 3 times, but nothing changes here. My computer is behind corporate proxy and cannot access the signature server from here. When signing from my mobile, it does not seem to have any effect.

Just make sure your local github user has the same email address set you used to sign the CLA

PR #483, opened 2022-10-30
chore: remove android.permission.WRITE_EXTERNAL_STORAGE

KazuCocoa:
Afaik, this external storage permission existed from the beginning. Espresso driver does not have this, so potentially can uia2 server also remove this...? This may needed something old devices such as SD card required one to save data...?

I tested with this change in ruby_lib_core's test cases locally, including push/pull files and screenshot stuff, but no error occurred related to this permission.

(This is not https://github.com/appium/appium/issues/17679 's fix. I just thought this permission's necessity)

updated:

Perhaps.... push to SD card storage area may require this...? <- push/pull is basically via adb, so maybe not needed for the permission

> In particular, if your app targets Android 11 (API level 30) or higher, the WRITE_EXTERNAL_STORAGE permission doesn't have any effect on your app's access to storage.

https://developer.android.com/training/data-storage
Current `targetSdkVersion` is 30, so the permission no longer have effort anyway...?

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -3,7 +3,6 @@
 
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.CHANGE_WIFI_STATE" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,6 +1,13 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           xmlns:tools="http://schemas.android.com/tools" package="io.appium.uiautomator2.test">
 
+    <!-- Note:
+    android.permission.WRITE_EXTERNAL_STORAGE and android.permission.READ_EXTERNAL_STORAGE
+    could cause the server process is killed since Android 6 as permission changes.
+    We should avoid using them as possible to prevent the behavior.
+    https://github.com/appium/appium/issues/17679
+    -->
+
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />


KazuCocoa:
Another description from the data-storage page:

> On earlier versions of Android, apps needed to declare the READ_EXTERNAL_STORAGE permission to access any file outside the [app-specific directories](https://developer.android.com/training/data-storage/app-specific#external) on external storage. Also, apps needed to declare the WRITE_EXTERNAL_STORAGE permission to write to any file outside the app-specific directory.

app/src/main/AndroidManifest.xml
```diff
@@ -3,7 +3,6 @@
 
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
```
mykola-mokhnach, line 6: Consider replacing it with READ_EXTERNAL_STORAGE one

keyur2maru:
@KazuCocoa @mykola-mokhnach I think this might be the fix for my [Issue](https://github.com/appium/appium/issues/17679) because of the following message in the logs: 
Invalidating LocalCallingIdentity cache for package io.appium.settings. Reason: op android:read_external_storage
I am using a DPC of a MDM to provision all the devices in the test that I am running. It might be the case that the DPC is revoking permissions on the devices where the test failed and on others where it passes it may be the case that the test passes before it gets to revoke the permissions.

app/src/main/AndroidManifest.xml
KazuCocoa: Does the WRITE_ include the READ_ permission...? I thought they were different permission, and they are basically does not required since api level 19. This apk now has minSdkVersion 21, so... perhaps we can just remote this.

READ_EXTERNAL_STORAGE
> Also starting in API level 19, this permission is not required to read or write files in your application-specific directories returned by [Context.getExternalFilesDir(String)](https://developer.android.com/reference/android/content/Context#getExternalFilesDir(java.lang.String)) and [Context.getExternalCacheDir()](https://developer.android.com/reference/android/content/Context#getExternalCacheDir()).

WRITE_EXTERNAL_STORAGE
> Starting in API level 19, this permission is not required to read/write files in your application-specific directories returned by [Context.getExternalFilesDir(String)](https://developer.android.com/reference/android/content/Context#getExternalFilesDir(java.lang.String)) and [Context.getExternalCacheDir()](https://developer.android.com/reference/android/content/Context#getExternalCacheDir()).

KazuCocoa:
> io.appium.settings. Reason: op android:read_external_storage

The package is https://github.com/appium/io.appium.settings , not this one.

>  It might be the case that the DPC is revoking permissions on the devices

Hm.. then, in your environment, other permissions, each Appium's apk has, also causes the reported issue, I guess. It is by system which is out of Appium's control.

keyur2maru:
> > io.appium.settings. Reason: op android:read_external_storage
> 
> The package is https://github.com/appium/io.appium.settings , not this one.
> 
> > It might be the case that the DPC is revoking permissions on the devices
> 
> Hm.. then, in your environment, other permissions, each Appium's apk has, also causes the reported issue, I guess. It is by system which is out of Appium's control.

I also found this in same log
Invalidating LocalCallingIdentity cache for package io.appium.uiautomator2.server. Reason: op android:read_external_storage

Does Appium server uses pre-built apks when I launch a session on the server? If yes, please share the location of the prebuilt apks for io.appium.settings and io.appium.uiautomator2.server which the server uses. I will try removing WRITE_EXTERNAL_STORAGE and READ_EXTERNAL_STORAGE permissions and see if it helps resolving my issue

mykola-mokhnach (approved):

PR #487, opened 2022-11-08
refactor: Reset session-specific settings before a new session starts

mykola-mokhnach:
Related to https://github.com/appium/appium/pull/17727


jlipps (approved):
makes sense. i'll update the doc to explain that there might be some settings which outlast a session.

PR #502, opened 2023-03-04
feat: get installed packages information

shashank2086:
Getting all Installed and System apps of your device with Name and Main Activity for easily selecting and starting automation of any Application.


mykola-mokhnach:
What is the purpose of this extension? Packages querying could be easily implemented on the driver level via ADB.

shashank2086:
Well with Adb it only gives me package name of the application, Like for my requirement, i wanted to get more information like the app name for easy selection and The Activity name which i need to for creating a session. And also if appium server is giving me directly the app list then i don't need to run the terminal command.

mykola-mokhnach:
> Adb it only gives me package name of the application

Probably you just need to use a different command then. Generally, `adb shell pm` can fetch all needed information about installed packages.

shashank2086:
I have tried alot but couldn't find the commands to get exact information i wanted that's why i wrote this enhancement, I think it's more useful and efficient if the appium server is able to get me these informations as well.

mykola-mokhnach:
android.permission.QUERY_ALL_PACKAGES permission has only been introduced in Android R. It won't work with older Android versions.

The pm utility though will work.

app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
```diff
@@ -40,12 +40,13 @@
...
 import static io.appium.uiautomator2.utils.Device.getUiDevice;
 
+
 public class ServerInstrumentation {
     private static final int MIN_PORT = 1024;
     private static final int MAX_PORT = 65535;
     private static final String WAKE_LOCK_TAG = "UiAutomator2:ScreenKeeper";
     public static final long MAX_TEST_DURATION = 24 * 60 * 60 * 1000;
-
+    public static Context ServerContext = null;
```
mykola-mokhnach, line 49: the context must be retrieved from the Instrumentation singleton

KazuCocoa:
`adb shell dumpsys activity activities` and `adb shell dumpsys package <packagename>` as well? Perhaps they will give more information though.

I'm curious about what information could get only via this

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
```diff
@@ -161,6 +162,7 @@ private void registerPostHandler() {
 
     private void registerGetHandler() {
         register(getHandler, new Status("/status"));
+        register(getHandler,new Getpackages("/apps"));
```
mykola-mokhnach, line 165: this is not a standard w3c endpoint, so we should not expose it like this. Consider adding this data into the existing /info endpoint instead

app/src/main/java/io/appium/uiautomator2/model/api/touch/PackageModel.java
```diff
@@ -0,0 +1,31 @@
+package io.appium.uiautomator2.model.api.touch;
```
mykola-mokhnach, line 1: the standard copyright header is missing

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
```diff
@@ -0,0 +1,47 @@
+package io.appium.uiautomator2.handler;
```
mykola-mokhnach, line 1: the standard copyright header is missing

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
```diff
@@ -0,0 +1,47 @@
...
+import io.appium.uiautomator2.handler.request.SafeRequestHandler;
+import io.appium.uiautomator2.http.AppiumResponse;
+import io.appium.uiautomator2.http.IHttpRequest;
+import io.appium.uiautomator2.model.api.BatteryStatusModel;
+import io.appium.uiautomator2.model.api.StatusModel;
+import io.appium.uiautomator2.model.api.touch.PackageModel;
+import io.appium.uiautomator2.server.ServerInstrumentation;
+import io.appium.uiautomator2.utils.BatteryHelper;
+
+public class Getpackages extends SafeRequestHandler implements NoSessionCommandHandler {
```
mykola-mokhnach, line 24: GetPackages

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
```diff
@@ -0,0 +1,47 @@
+package io.appium.uiautomator2.handler;
+
+
+import static io.appium.uiautomator2.model.Session.NO_ID;
```
mykola-mokhnach, line 4: Please format the code same way all other modules are formatted

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
```diff
@@ -0,0 +1,47 @@
...
+import io.appium.uiautomator2.model.api.StatusModel;
+import io.appium.uiautomator2.model.api.touch.PackageModel;
+import io.appium.uiautomator2.server.ServerInstrumentation;
+import io.appium.uiautomator2.utils.BatteryHelper;
+
+public class Getpackages extends SafeRequestHandler implements NoSessionCommandHandler {
+    public Getpackages(String mappedUri) {
+        super(mappedUri);
+    }
+    public static List<PackageModel> appDetails = new ArrayList<PackageModel>();
```
mykola-mokhnach, line 28: please move this pvoiperty to the method

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
```diff
@@ -0,0 +1,47 @@
...
+            for(ApplicationInfo appInfo: apps) {
+                PackageModel model = new PackageModel();
+                if(ServerContext.getPackageManager().getLaunchIntentForPackage(appInfo.packageName)!=null) {
+                    model.setPackageName(appInfo.packageName);
+                    model.setAppName((String) ServerContext.getPackageManager().getApplicationLabel(appInfo));
+                    model.setPackageActivity(ServerContext.getPackageManager().getLaunchIntentForPackage(appInfo.packageName).getComponent().getClassName());
+                    appDetails.add(model);
+                }
+            }
+        }catch (Exception e){
```
mykola-mokhnach, line 42: is it really necessary to silence exceptions? where and which exceptions could happen there?

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    protected AppiumResponse safeHandle(IHttpRequest request) {
+        try{
+            List<ApplicationInfo> apps = ServerContext.getPackageManager().getInstalledApplications(PackageManager.GET_META_DATA);
+            for(ApplicationInfo appInfo: apps) {
+                PackageModel model = new PackageModel();
+                if(ServerContext.getPackageManager().getLaunchIntentForPackage(appInfo.packageName)!=null) {
+                    model.setPackageName(appInfo.packageName);
+                    model.setAppName((String) ServerContext.getPackageManager().getApplicationLabel(appInfo));
+                    model.setPackageActivity(ServerContext.getPackageManager().getLaunchIntentForPackage(appInfo.packageName).getComponent().getClassName());
```
mykola-mokhnach, line 38: this line is too long

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
```diff
@@ -0,0 +1,47 @@
...
+import io.appium.uiautomator2.http.AppiumResponse;
+import io.appium.uiautomator2.http.IHttpRequest;
+import io.appium.uiautomator2.model.api.BatteryStatusModel;
+import io.appium.uiautomator2.model.api.StatusModel;
+import io.appium.uiautomator2.model.api.touch.PackageModel;
+import io.appium.uiautomator2.server.ServerInstrumentation;
+import io.appium.uiautomator2.utils.BatteryHelper;
+
+public class Getpackages extends SafeRequestHandler implements NoSessionCommandHandler {
+    public Getpackages(String mappedUri) {
```
mykola-mokhnach, line 25: empty constructor declaration could be skipped

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
```diff
@@ -0,0 +1,47 @@
...
+        super(mappedUri);
+    }
+    public static List<PackageModel> appDetails = new ArrayList<PackageModel>();
+    @Override
+    protected AppiumResponse safeHandle(IHttpRequest request) {
+        try{
+            List<ApplicationInfo> apps = ServerContext.getPackageManager().getInstalledApplications(PackageManager.GET_META_DATA);
+            for(ApplicationInfo appInfo: apps) {
+                PackageModel model = new PackageModel();
+                if(ServerContext.getPackageManager().getLaunchIntentForPackage(appInfo.packageName)!=null) {
```
mykola-mokhnach, line 35: the package manager instance should be retrieved once.

mykola-mokhnach:
Please assign a proper name to the pull requests according to  https://www.conventionalcommits.org requirements

app/src/main/java/io/appium/uiautomator2/model/api/touch/PackageModel.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class PackageModel {
+    String PackageName;
```
mykola-mokhnach, line 4: member names should be in `camelCase`

mykola-mokhnach: you could also make them `private final` as they are not supposed to be mutable during model lifecycle

app/src/main/java/io/appium/uiautomator2/model/api/touch/PackageModel.java
```diff
@@ -0,0 +1,31 @@
...
+    String PackageActivity;
+    String AppName;
+
+    public String getAppName() {
+        return AppName;
+    }
+
+    public String getPackageName() {
```
mykola-mokhnach, line 12: Can any of the values be `null`? If yes then mark corresponding getters as `@Nullable`

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
mykola-mokhnach: you could simply use the built-in Android Studio format feature (Cmd+Shift+L)

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
mykola-mokhnach: the current implementation will end up with all packages list duplicated there as many times this endpoint is called

app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
mykola-mokhnach: always try to avoid making public properties mutable, except of pure data classes.

shashank2086: You mean i should create a method to retrieve the Context Right?

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
shashank2086: For driver, we have to add in the driver Wrapper SDK right?

shashank2086: I have changed the endpoint to /appium/device/apps

shashank2086:
> activity

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
shashank2086: Yes possibly, i will do the changes thanks

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
shashank2086: > `adb shell dumpsys activity activities` and `adb shell dumpsys package <packagename>` as well? Perhaps they will give more information though.
> 
> I'm curious about what information could get only via this

Well first command will only give information of the activities information of the current opened application, and also the main issue is reading the dumpsys is another headache, maybe we can achieve but this seems the more efficient way of doing

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
KazuCocoa: The return value description from the documentation of `getLaunchIntentForPackage` is:

> The current implementation looks first for a main activity in the category [Intent#CATEGORY_INFO](https://developer.android.com/reference/android/content/Intent#CATEGORY_INFO), and next for a main activity in the category [Intent#CATEGORY_LAUNCHER](https://developer.android.com/reference/android/content/Intent#CATEGORY_LAUNCHER).

So, maybe the name should be `launcherActivityName` or similar to indicate it is just for possible launcher activity.

Btw, have you tried `getLaunchIntentSenderForPackage`?  https://developer.android.com/reference/android/content/pm/PackageManager#getLaunchIntentSenderForPackage(java.lang.String)
I wonder if the api is proper for android sdk level 33+

shashank2086: Thanks for the suggestion, let me check, i will also try to check in almost all SDKs using emulators

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
shashank2086: Yes exactly, So is packageName and activityName correct? or should i rename to any other name for better understanding?

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
shashank2086: I have tried <getLaunchIntentSenderForPackage> but it's not giving me the component info but <getLaunchIntentForPackage> working perfectly

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
```diff
@@ -0,0 +1,57 @@
...
+import io.appium.uiautomator2.utils.Logger;
+
+public class GetPackages extends SafeRequestHandler implements NoSessionCommandHandler {
+    public GetPackages(String mappedUri) {
+        super(mappedUri);
+    }
+    @Override
+    protected AppiumResponse safeHandle(IHttpRequest request) {
+        List<PackageModel> appDetails = new ArrayList<PackageModel>();
+        try{
```
mykola-mokhnach, line 39: the formatting is still not fully addressed. Block parentheses must be separated with spaces

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
```diff
@@ -0,0 +1,57 @@
...
+            PackageManager manager = getApplicationContext().getPackageManager();
+            List<ApplicationInfo> apps = manager.getInstalledApplications(manager.GET_META_DATA);
+            for(ApplicationInfo appInfo: apps) {
+                PackageModel model = new PackageModel();
+                if(manager.getLaunchIntentForPackage(appInfo.packageName)!=null) {
```
mykola-mokhnach, line 44: why packages without launch intents must be excluded?

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
```diff
@@ -0,0 +1,57 @@
...
+                    model.setPackageName(appInfo.packageName);
+                    model.setAppName((String) manager.getApplicationLabel(appInfo));
+                    model.setPackageActivity(manager.getLaunchIntentForPackage(appInfo.packageName).getComponent().getClassName());
+                    appDetails.add(model);
+                }
+            }
+
+        }catch (RuntimeException e){
```
mykola-mokhnach, line 52: this exception handler is still to generic

app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
```diff
@@ -45,9 +45,9 @@ public class ServerInstrumentation {
     private static final int MAX_PORT = 65535;
     private static final String WAKE_LOCK_TAG = "UiAutomator2:ScreenKeeper";
     public static final long MAX_TEST_DURATION = 24 * 60 * 60 * 1000;
-
+
```
mykola-mokhnach, line 48: Please revert the changes

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
mykola-mokhnach: does it make sense to have this endpoint without a session id?

app/src/main/java/io/appium/uiautomator2/model/api/touch/PackageModel.java
mykola-mokhnach: this is still relevant. I don't see a reason to make model fields mutable

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
```diff
@@ -0,0 +1,57 @@
...
+        List<PackageModel> appDetails = new ArrayList<PackageModel>();
+        try{
+            PackageManager manager = getApplicationContext().getPackageManager();
+            List<ApplicationInfo> apps = manager.getInstalledApplications(manager.GET_META_DATA);
+            for(ApplicationInfo appInfo: apps) {
+                PackageModel model = new PackageModel();
+                if(manager.getLaunchIntentForPackage(appInfo.packageName)!=null) {
+                    model.setPackageName(appInfo.packageName);
+                    model.setAppName((String) manager.getApplicationLabel(appInfo));
+                    model.setPackageActivity(manager.getLaunchIntentForPackage(appInfo.packageName).getComponent().getClassName());
```
mykola-mokhnach, line 47: the line iss too long

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
mykola-mokhnach: > For driver, we have to add in the driver Wrapper SDK right?

There should be a driver endpoint, which would expose this information. You could implement it in two ways, either using a `mobile: ...` extension or by adding a custom API endpoint and assume that clients would add a support for it.

mykola-mokhnach: Example of a mobile extension: https://github.com/appium/appium-uiautomator2-driver/blob/7013ba401252a7de59c2fe15516440db4e606a4c/lib/commands/battery.js#L24

Example of a custom API endpoint: https://github.com/appium/appium-uiautomator2-driver/blob/7013ba401252a7de59c2fe15516440db4e606a4c/lib/method-map.js#L5

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
```diff
@@ -0,0 +1,57 @@
...
+public class GetPackages extends SafeRequestHandler implements NoSessionCommandHandler {
+    public GetPackages(String mappedUri) {
+        super(mappedUri);
+    }
+    @Override
+    protected AppiumResponse safeHandle(IHttpRequest request) {
+        List<PackageModel> appDetails = new ArrayList<PackageModel>();
+        try{
+            PackageManager manager = getApplicationContext().getPackageManager();
+            List<ApplicationInfo> apps = manager.getInstalledApplications(manager.GET_META_DATA);
```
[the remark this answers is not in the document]
shashank2086: It can't be null but if any chance it dint get the apps, it will give back empty data

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: Well it was also giving unnecessary sub pacakges without intent, to filter out the main system and installed applications i have provided the condition

app/src/main/java/io/appium/uiautomator2/server/ServerInstrumentation.java
shashank2086: Sure

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: Done can you check now

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: done

app/src/main/java/io/appium/uiautomator2/model/api/touch/PackageModel.java
shashank2086: Done please verify

app/src/main/java/io/appium/uiautomator2/model/api/touch/appium/PackageModel.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.uiautomator2.model.api.touch.appium;
+
+import javax.annotation.Nullable;
+
+public class PackageModel {
+    private final @Nullable
```
mykola-mokhnach, line 22: getters could still be preserved and have the necessary annotation

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
mykola-mokhnach: Please add this as an explanation comment

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
mykola-mokhnach: My question is more why/whether this is needed? Shouldn't it be a valid endpoint error if any of these calls fails?

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: okay got it

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: I guess it's not needed, since all the returning properties are nullable and doesn't throws any Exception except getApplicationLabel but it throws when the activity is not cannot be found and we already have condition above to check that. So should i remove it?

app/src/main/java/io/appium/uiautomator2/model/api/touch/appium/PackageModel.java
shashank2086: okay got it

shashank2086: done

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: done

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
mykola-mokhnach: yes, it is better to remove it

app/src/main/java/io/appium/uiautomator2/handler/Getpackages.java
KazuCocoa: thanks for the checking

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    protected AppiumResponse safeHandle(IHttpRequest request) {
+        List<PackageModel> appDetails = new ArrayList<PackageModel>();
+        PackageManager manager = getApplicationContext().getPackageManager();
+        List<ApplicationInfo> apps = manager.getInstalledApplications(manager.GET_META_DATA);
+        for (ApplicationInfo appInfo : apps) {
+            // Filtering out unnecessary sub packages without Intent
+            if (manager.getLaunchIntentForPackage(appInfo.packageName) != null) {
+                appDetails.add(new PackageModel(appInfo.packageName,
+                        manager.getLaunchIntentForPackage(appInfo.packageName).getComponent().getClassName(),
```
KazuCocoa, line 48: What happen when `getLaunchIntentForPackage(appInfo.packageName)` is null, or `getComponent` is null. They are nullable.

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: done

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: They both are Nullable, if an application has an intent then for sure it will have a component, so application with no component will automatically get filter out from the 46 condition and also getComponent is also Nullable

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
shashank2086: Got it, i will do the necessary changes

shashank2086: > does it make sense to have this endpoint without a session id?

Yes before starting a session, I should be able to get the application list for easy selection to start session on any of the listed application

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
KazuCocoa: ah, missed here called the method twice.
This endpoint tries to show installed packages and each information, then maybe if would be nice to keep the package name (since it exists) but no activities rather than filtering the package out entirely. What do you think?
So, the user can understand some packages exist in the device but they has no launchable activities by this endpoint.

shashank2086: The intend was to show all the main applications since the package without Intent is just a subpackage of it's parent package which can just create confusion to the user and actually it's pretty much useless, if later on anyone needs that then surely we can modify accordingly but i feel it's unnecessary right now to show?

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
mykola-mokhnach: One cannot call any endpoint without having a session id. This is just how selenium protocol works. You need to call createSession first, get a unique id and then use it to call other APIs. No endpoint could be called prior to createSession.

So, how/where this implementation is supposed to be used by UIA2 driver?

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
```diff
@@ -0,0 +1,54 @@
...
+import io.appium.uiautomator2.model.api.touch.appium.PackageModel;
+
+public class GetPackages extends SafeRequestHandler implements NoSessionCommandHandler {
+    public GetPackages(String mappedUri) {
+        super(mappedUri);
+    }
+
+    @Override
+    protected AppiumResponse safeHandle(IHttpRequest request) {
+        List<PackageModel> appDetails = new ArrayList<PackageModel>();
```
mykola-mokhnach, line 41: could be simplified to `new ArrayList<>()`

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
```diff
@@ -0,0 +1,54 @@
...
+        PackageManager manager = getApplicationContext().getPackageManager();
+        List<ApplicationInfo> apps = manager.getInstalledApplications(manager.GET_META_DATA);
+        for (ApplicationInfo appInfo : apps) {
+            // Filtering out unnecessary sub packages without Intent
+            if (manager.getLaunchIntentForPackage(appInfo.packageName) != null) {
+                appDetails.add(new PackageModel(appInfo.packageName,
+                        manager.getLaunchIntentForPackage(appInfo.packageName).getComponent().getClassName(),
+                        (String) manager.getApplicationLabel(appInfo)));
```
mykola-mokhnach, line 49: Use charSequenceToNullableString or charSequenceToString helpers from utils.StringHelpers pcakage

app/src/main/java/io/appium/uiautomator2/model/api/touch/appium/PackageModel.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package io.appium.uiautomator2.model.api.touch.appium;
+
+import javax.annotation.Nullable;
+
+public class PackageModel {
+    private final
+    String packageName;
+    private final @Nullable
```
mykola-mokhnach, line 24: @Nullable is redundant

app/src/main/java/io/appium/uiautomator2/model/api/touch/appium/PackageModel.java
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.uiautomator2.model.api.touch.appium;
+
+import javax.annotation.Nullable;
+
+public class PackageModel {
+    private final
```
mykola-mokhnach, line 22: please join property definitions into a single string, e.g.: `private final String packageName;`

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: Okay

app/src/main/java/io/appium/uiautomator2/model/api/touch/appium/PackageModel.java
shashank2086: Okay

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: Sorry You are talking about casting?

mykola-mokhnach: yes

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
shashank2086: Correct, with driver we have to create session to utilize any api, from driver user has to create session to utilize the this API, i am utilizing the apis directly so i don't need to create session

shashank2086: So if user is using diver than he can utilize after creating session

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: done

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: done

app/src/main/java/io/appium/uiautomator2/model/api/touch/appium/PackageModel.java
shashank2086: done

app/src/main/java/io/appium/uiautomator2/model/api/touch/appium/PackageModel.java
shashank2086: done

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
mykola-mokhnach: I'm not quite convinced of adding an API, which is not supposed to be used by UIA2 driver then. You should fork the project if you use it in a custom way.

@KazuCocoa what do you think?

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
KazuCocoa: Then, could you leave the intention as docstring in this class for future maintenance?
`GetPackages` made me think the app package is primary, then app activities etc as additional information. The current intention is to get app activities information than its package (since filter by package).

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
KazuCocoa: I'm ok to add a new endpoint, if not a complicated feature to keep maintaining, but I'd request to add test code to check the function. Endpoints which are not in regular WebDriver protocol and not used by UIA2 driver regularly could break because we don't care so much for as our project. (Actually, fork is the most flexible way to maintain your own way as Mykola addressed.)

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
KazuCocoa: Yea, could you leave what kind of activities will be available, and no activities packages won't be available?
I mean not a comment for the implementation, just this method's entire behavior as the module's description. Also, it wole be appreciate to add this method uses `android.permission.QUERY_ALL_PACKAGES` here, or a comment in the manifest to link each

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
shashank2086: Well i can maintain it, but i think it's a quite a useful feature for the End User to have and i think it can be easily maintained and should not affect the existing implementation

shashank2086:
Can it be merged now?

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: done

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
mykola-mokhnach: I have already commented https://github.com/appium/appium-uiautomator2-driver/pull/585
It is just not going to work without a session id.

If you really need to have the endpoint in this rep I can propose to create two handlers for it - one with session id and one without (`session/:sessionId/appium/device/apps` and `/appium/device/apps` ). The latter could be then used in custom calls, but since the code would be shared, we can guarantee it works as expected.

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
```diff
@@ -0,0 +1,56 @@
...
+
+import io.appium.uiautomator2.handler.request.NoSessionCommandHandler;
+import io.appium.uiautomator2.handler.request.SafeRequestHandler;
+import io.appium.uiautomator2.http.AppiumResponse;
+import io.appium.uiautomator2.http.IHttpRequest;
+import io.appium.uiautomator2.model.api.touch.appium.PackageModel;
+
+
+public class GetPackages extends SafeRequestHandler implements NoSessionCommandHandler {
+    public GetPackages(String mappedUri) {
```
mykola-mokhnach, line 37: It would make sense to add at least one integration test for this endpoint

app/src/main/java/io/appium/uiautomator2/server/AppiumServlet.java
shashank2086: Okay makes sense, i will do that

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
KazuCocoa: Sorry, missing commit?

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: I have added

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
shashank2086: Done now

app/src/androidTestE2eTest/java/io/appium/uiautomator2/unittest/test/DeviceCommandsTest.java
```diff
@@ -824,4 +825,11 @@ public void shouldExtractDeviceInformation() throws JSONException {
         assertEquals(info.getString("platformVersion"), Build.VERSION.RELEASE);
         assertTrue(info.getJSONArray("networks").length() > 0);
     }
+
+    @Test
+    public void shouldExtractDeviceApplications() throws JSONException {
```
mykola-mokhnach, line 830: the terminology used is a bit confusing to me
ExtractDeviceApplications vs getListApplications vs GetPackages
is it possible to follow the same terminology everywhere?

shashank2086: Done the formatting

mykola-mokhnach (approved):

app/src/main/java/io/appium/uiautomator2/handler/GetPackages.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import io.appium.uiautomator2.handler.request.NoSessionCommandHandler;
+import io.appium.uiautomator2.handler.request.SafeRequestHandler;
+import io.appium.uiautomator2.http.AppiumResponse;
+import io.appium.uiautomator2.http.IHttpRequest;
+import io.appium.uiautomator2.model.api.touch.appium.PackageModel;
+
+// Get list of all applications (System and Installed) with packageActivity
```
KazuCocoa, line 36:
```suggestion
// Get a list of all applications (System and Installed) with packageActivity.
// It does not include applications wich has no launchable activities in the manifest.
```

KazuCocoa (approved):

mykola-mokhnach:
@KazuCocoa Do you have any additional comments?

KazuCocoa:
no, just waited for the ci result

PR #509, opened 2023-05-27
fix: Revert "feat: settings API to get the topmost window from active package"

mykola-mokhnach:
Unfortunately it turns out Optionals and streams only work since API 24, although the server still supports API21+

It is necessary to rewrite the code, so it uses vanilla loops and no optionals in order to make it compatible with older Android APIs.


mykola-mokhnach:
cc @yaumu3

KazuCocoa (approved):

yaumu3:
Thanks for letting me know. I'll submit fixups.

PR #510, opened 2023-05-28
feat: Add a possibility of scheduling and unscheduling actions

mykola-mokhnach:
The idea of this PR is to improve the handling of shortly-living UI elements on the client side. 

Webdriver protocol is REST-based, which means each particular command must be received from a remote client, processed by the server and then replied. It gets even longer in case we have proxies in between, which is always the case for cloud providers. This whole process takes time, which might be critical if the script tries to handle short-living UI primitives, like popups. Basically, a popup might already disappear until the UIA2 server gets a command to click it if the TTL of the popup is 3 seconds and the comm between server and client takes more than that.

Now about how this PR approaches the above issue:

If we known that the popup will appear later then on the client side we call something like

```python
driver.execute_script('mobile: scheduleAction', {
  'name': 'myPopupHandlingAction',
  'steps': [{
    'type': 'gesture',
    'name': 'clickPopupButtonStep',
    'payload': {
      'subtype': 'click',
      'locator': {
        'strategy': 'id',
        'selector': 'buttonIdentifier',
      }
    }
  }],
  'maxPass': 1,
  'times': 20,
})
```

which asks the server to *asynchronously* try the click action on the element located by `buttonIdentifier` every second up to 20 times. The script is also asked to stop as soon as at least one click passes.

Afterwards the client may asynchronously request to fetch the action history or to unschedule it explicitly.

Currently I have implemented 3 subtypes of `gesture` actions (click, doubleClick, and longClick) and 1 subtype of `source` action (xml).

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/handler/DeleteSession.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/DeleteSession.java
@@ -26,6 +26,7 @@ import io.appium.uiautomator2.model.AppiumUIA2Driver;
 import io.appium.uiautomator2.model.NotificationListener;
 import io.appium.uiautomator2.model.Session;
 import io.appium.uiautomator2.server.ServerInstrumentation;
+import io.appium.uiautomator2.utils.actions_scheduler.ScheduledActionsManager;
 
 public class DeleteSession extends SafeRequestHandler {
 
@@ -40,6 +41,7 @@ public class DeleteSession extends SafeRequestHandler {
         if (currentSession == null || !Objects.equals(sessionId, currentSession.getSessionId())) {
             throw new NoSuchDriverException(String.format("The session %s cannot be found", sessionId));
         }
+        ScheduledActionsManager.getInstance().clear();
         NotificationListener.getInstance().stop();
         ServerInstrumentation.getInstance().stopServer();
         return new AppiumResponse(sessionId);
--- a/app/src/main/java/io/appium/uiautomator2/handler/FindElement.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/FindElement.java
@@ -16,11 +16,9 @@
 
 package io.appium.uiautomator2.handler;
 
-import androidx.annotation.Nullable;
 import androidx.test.uiautomator.UiObjectNotFoundException;
 
 import io.appium.uiautomator2.common.exceptions.ElementNotFoundException;
-import io.appium.uiautomator2.common.exceptions.NotImplementedException;
 import io.appium.uiautomator2.handler.request.SafeRequestHandler;
 import io.appium.uiautomator2.http.AppiumResponse;
 import io.appium.uiautomator2.http.IHttpRequest;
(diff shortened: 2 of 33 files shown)


KazuCocoa:
It sounds like http://appium.io/docs/en/2.0/reference/commands/execute-driver-plugin/ ...?

https://github.com/appium/python-client/blob/5a4b6d0729601b3fdb821e93ebe24bf2d248c65b/test/unit/webdriver/execute_driver_test.py#L36-L41

So, the execute_driver block does find element -> does action (e.g. click) -> return the response

mykola-mokhnach:
yes, it is somehow similar to that one, but has several important pros:
- does not require special security permssions
- does not require javascript knowledge to write the actual code
- is executed asynchronously, e.g. the schedule API does not block, so one can continue sending commands to the driver while the original block is still being executed

KazuCocoa:
i see, sounds reasonable to add a new command.

Can it add error handling to the argument?

Current expected behavior of `subtype` is a single action for ok case. In the 3 seconds TTL case, maybe checking the status in the error case also important. So I wondered if it could have error handling like `onError` actions to take a screenshot and get the source, for example, if the given action gets failed.

mykola-mokhnach:
> Current expected behavior of `subtype` is a single action for ok case. In the 3 seconds TTL case, maybe checking the status in the error case also important. So I wondered if it could have error handling like `onError` actions to take a screenshot and get the source in the 3 seconds limit, for example, if the given action gets failed. (the `times` also affect the timing tho)

I find callbacks might be too complicated to define. What we could add is probably an option to add a page source and/or a screenshot to failed actions. But don't forget that such actions on their own need time.
The original idea was that one could run an action that gathers the source periodically to catch the actual popup source and then just run `click` or whatever in order to handle it after the debug is completed. This would not seriously affect the general performance.

KazuCocoa:
Gotcha, sounds reasonable to not have such onError to me

KazuCocoa (approved):
the code lgtm with green ci results

jlipps:
I don't think I have a good idea of how generic this is intended to be, or what the semantics of the different keys are. Would it be possible to write up the user facing documentation so I can get a better idea of what this API is supposed to encompass? In general I think it overlaps a lot with Execute Driver Script.

mykola-mokhnach:
> I don't think I have a good idea of how generic this is intended to be, or what the semantics of the different keys are. Would it be possible to write up the user facing documentation so I can get a better idea of what this API is supposed to encompass? In general I think it overlaps a lot with Execute Driver Script.

Here you have it @jlipps: https://github.com/appium/appium-uiautomator2-driver/pull/616

mykola-mokhnach:
> thanks, very helpful @mykola-mokhnach. So the way I see it these steps are "blind", that is, there is no possibility of having one step conditional on another step or using information from a prior step. This feature basically lets you swing a hammer in the background. Got it. I am curious to gather more examples of its usage beyond handling popups. And I have the feeling that users will want more control. But I guess they can just use Execute Driver Script if they want to do fancier things.

In general yes. It is just that not everyone knows how to code wdio and/or not all cloud providers may allow this feature at all.

PR #508, opened 2023-05-26
feat: settings API to get the topmost window from active package

yaumu3:
## Motivation

This Pull Request presents a feature that enables the retrieval of a non-focusable window from the same package, positioned at a layer above the active window. Similar to #301.

This enhancement offers an additional advantage (especially when using XPath locators) in E2E tests. Unlike the `enableMultiWindows` setting, this new feature negates the need to consider the number of windows displayed at any given time, thereby improving test portability.

## What I change

- Implementation of a new strategy to identify and retrieve the topmost window among all windows corresponds to the active package.
- Addition of `enableTopmostWindowFromActivePackage` settings API to toggle this feature.
- By default, the feature is disabled, ensuring that existing functionalities remain unaffected unless the user explicitly enables it.

## What I checked

- A manual checked was carried out using [Appium Inspector](https://github.com/appium/appium-inspector) both when the option was enabled and disabled.
- All unit tests, as outlined in the `README.md`, were successfully executed.

PR #512, opened 2023-05-31
feat: settings API to get the topmost window from active package

yaumu3:
This mirrors #508 but it addresses the compatibility issue reported in #509 by substituting code segments that are unsupported in certain older API levels.

The motivation and modifications introduced in #508 are also applicable to this one.

## What I checked (in addition to those in #508)

- Manually tested on the specified API levels (including the minimum supported one) to ensure that there are no errors upon launch.
    - [x] Android 5.0 (API level 21) 
    - [x] Android 12.0 (API level 31)
    - [x] Android 13.0 (API level 33)

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
+++ b/app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
@@ -123,7 +123,9 @@ public class AXWindowHelpers {
                 return root;
             }
         }
-        throw new UiAutomator2Exception("No active window root found");
+        throw new UiAutomator2Exception(
+                String.format("Unable to find the active topmost window associated with %s package",
+                        activeRootPackageName));
     }
 
     public static AccessibilityNodeInfo[] getCachedWindowRoots() {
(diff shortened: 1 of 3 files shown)


KazuCocoa (approved):

app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
```diff
@@ -99,20 +106,49 @@ private static AccessibilityNodeInfo[] getWindowRoots() {
...
+            }
+        });
+        for (AccessibilityWindowInfo window : windows) {
+            AccessibilityNodeInfo root = window.getRoot();
+            if (root != null &&
+                    Objects.equals(root.getPackageName(), activeRootPackageName)) {
+                return root;
+            }
+        }
+        throw new UiAutomator2Exception("No active window root found");
```
mykola-mokhnach, line 126: We could improve the error message like 

Cannot find the active topmost window belonging to xxx package

mykola-mokhnach (approved):

app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
yaumu3: I agree that it is understandable when the package name is shown. Revised the error message with 9cff0a99d9afbc96bff7b7b4e40dbd560fc594cf.

Thank you.

PR #518, opened 2023-06-29
fix: Ignore start coordinates for the very first move action

mykola-mokhnach:
We do not want to calculate coordinates for a very first pointerMove action in the chain as it is there to simply define the initial position of the cursor

Related to https://github.com/appium/java-client/issues/1954


KazuCocoa (approved):
seems reasonable

PR #525, opened 2023-08-18
chore : remove "/wd/hub" prefix to improve compatibility

howard1005:
This PR aims to enhance compatibility with previous appium versions by removing the "/wd/hub" prefix from URIs.

I am currently using Appium version 1.22.3, and I faced compatibility issues with the latest version of UIAutomator2 regarding the URI. I discovered that requests with the prefix "/wd/hub" were not compatible. By implementing a change to remove this specific prefix from incoming requests, I was able to enhance compatibility, allowing the latest version of UIAutomator2 to work with previous versions of Appium as well. This solution ensures a more seamless integration between these different versions, bridging the gap that existed due to the URI inconsistency.


jlipps:
Unfortunately compatibility with Appium 1 is not a goal for this driver anymore. It's only meant to be used with Appium 2.

PR #532, opened 2023-09-14
ci: remove npm update

KazuCocoa:
to avoid the latest npm's nodejs version mismatch error


.github/workflows/unit-test.yml
```diff
@@ -33,8 +33,6 @@ jobs:
     - uses: actions/setup-node@v3
       with:
         node-version: ${{ matrix.node-version }}
-    - run: npm i -g npm
```
mykola-mokhnach, line 36: we could also update the matrix generation algorithm above

PR #582, opened 2023-12-25
feat: bump targetSdkVersion from 30 to 31, and to 32

KazuCocoa:
Lets try to use newer `targetSdkVersion`

AS has helpful feature to show suggestions we should fix in the upgrade. According to the hint, we should:

Screenshot 2023-12-25 at 12 34 10 PM

Screenshot 2023-12-25 at 12 34 17 PM

Bluetooth is the manifest update.

The 2nd one is https://developer.android.com/topic/performance/app-hibernation :
> If your app targets Android 11 (API level 30) or higher, and the user doesn't [interact with your app](https://developer.android.com/topic/performance/app-hibernation#app-usage) for a few months, the system places your app in a hibernation state. The system optimizes for storage space instead of performance, and the system protects user data. This system behavior is similar to what occurs when the user manually force-stops your app from system settings.

so maybe not related to the uia2 server.

The 3rd one is may not for the case. Possibly our settings app needs some update though.

The 4th one https://developer.android.com/about/versions/12/behavior-changes-12#notification-trampolines

> To improve app performance and UX, apps that target Android 12 or higher can't start activities from [services](https://developer.android.com/guide/components/services) or [broadcast receivers](https://developer.android.com/guide/components/broadcasts) that are used as notification trampolines. In other words, after the user taps on a notification, or an [action button](https://developer.android.com/training/notify-user/build-notification#Actions) within the notification, your app cannot call [startActivity()](https://developer.android.com/reference/android/content/Context#startActivity(android.content.Intent)) inside of a service or broadcast receiver.

Maybe we do not have the usage to start UIA2 server from the notification regularly.

After upgrading to 31, checked 32. Then no diffs as below, so i have bumped to 32 as well.

Screenshot 2023-12-25 at 1 18 15 PM

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -14,11 +14,11 @@ apply plugin: 'com.android.application'
 apply plugin: 'de.mobilej.unmock'
 
 android {
-    compileSdk 30
+    compileSdk 31
     defaultConfig {
         applicationId 'io.appium.uiautomator2'
         minSdkVersion 21
-        targetSdkVersion 30
+        targetSdkVersion 31
         versionCode 145
         archivesBaseName = 'appium-uiautomator2'
         /**
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -14,11 +14,11 @@ apply plugin: 'com.android.application'
 apply plugin: 'de.mobilej.unmock'
 
 android {
-    compileSdk 31
+    compileSdk 32
     defaultConfig {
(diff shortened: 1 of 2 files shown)


KazuCocoa:
i have manually checked the build on my local with android 11 and 12 real devices with simple get page source, find element and send keys

mykola-mokhnach (approved):

PR #584, opened 2023-12-26
ci:  add android 13

KazuCocoa:
lets see how this will be

### DIFF
--- a/.github/workflows/functional-test.yml
+++ b/.github/workflows/functional-test.yml
@@ -15,18 +15,23 @@ jobs:
         - platformVersion: "13.0"
           apiLevel: 33
           emuTag: google_apis
+          arch: x86_64
         - platformVersion: "11.0"
           apiLevel: 30
           emuTag: google_apis
+          arch: x86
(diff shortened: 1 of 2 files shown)


KazuCocoa:
toast timing and below should be updated for the newer emulator:

```
    @Test
    public void verifyDismissingAnAlertWithButton() throws JSONException {
        if (Build.VERSION.SDK_INT < 23) {
            return;
        }

        setupView();

        Response response = findElement(By.accessibilityId("OK Cancel dialog with a long message"));
        clickAndWaitForStaleness(response.getElementId());

//        response = dismissAlert("CANCEL");
        response = dismissAlert("Cancel");
        assertTrue(response.isSuccessful());
    }
```

mykola-mokhnach:
AFAIK integration tests on newer platform versions were never stable. Android emulators  running recent OS versions are quite resource demanding and weak GH VMs cannot ensure the necessary performance thus making tests flaky.
Perhaps, moving to a higher (paid) teer would help, I remember @jlipps has once mentioned such possibility

PR #587, opened 2023-12-30
feat: bump targetSdkVersion to 34

KazuCocoa:
Once concern was regex related, but in my testing with this build on android 14 with page source, find element with XPath and send keys did not have any errors. Maybe should be ok. Lets see how CI will be...

Screenshot 2023-12-29 at 7 32 40 PM

Screenshot 2023-12-29 at 7 32 48 PM

Screenshot 2023-12-29 at 7 32 54 PM

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -14,11 +14,11 @@ apply plugin: 'com.android.application'
 apply plugin: 'de.mobilej.unmock'
 
 android {
-    compileSdk 33
+    compileSdk 34
     defaultConfig {
         applicationId 'io.appium.uiautomator2'
         minSdkVersion 21
-        targetSdkVersion 33
+        targetSdkVersion 34
         versionCode 148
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
Maybe we should let the server to run for a while with the current changes before bumping it again? Otherwise it might be hard to find a failure reason if too many changes are published within a short period of time.

KazuCocoa:
Possibly yes.

So far, I assume with this change for Android 14 is the most biggest than previous updates especially non-sdk interface changes. https://developer.android.com/about/versions/14/changes/non-sdk-14 has csv file for the non-SDK interface in Android 14. So Android 14 with this target version update could cause some potential issues I haven't found in major usage such as page source, device info, find element etc in my quick testing.

KazuCocoa:
https://github.com/appium/appium-uiautomator2-server/pull/591 included this change

Issue #596, opened 2024-01-10
Caching appears to be partially broken since 5.15.0


eglitise:
(will add more environment details later)

After updating to Appium `2.4.1` + UiAutomator2 driver `2.40.0`, I started seeing some strange test failures: some elements that were previously being found, were no longer being identified.
This is a fairly rare occurrence, and appears to affect Jetpack Compose structures, where, when the children of the compose root element are changed, the new children cannot be found. The app source does still contain them, and the Inspector can also identify them (although after one or more additional refreshes), so it appears that something may be wrong with caching.
I tested earlier UiA2 driver versions, and discovered that the problem first manifested in UiA2 server `5.15.0` (the previous `5.14.0`, which I installed manually, was still fine).

I could reproduce the problem on devices with Android 12 and 14, but not Android 9.
I also checked the Appium logs, but at the very least on the default logging level, there was nothing unusual, just a 404 response due to the element not being found. So some additional logging is probably needed.
Will investigate more...

KazuCocoa:
https://github.com/appium/appium-uiautomator2-server/pull/583 is possible relevant one.
The method update was by AS's recommendation change as the old method no longer worked over Android 13 (api level 33), `Build.VERSION_CODES.TIRAMISU`.

> with Android 12 and 14, but not Android 9.

Oh, wait. The new clearCache does not work for Android 12, the version keeps using the old method, so this is potentially the UIA library internal by Android SDK 33+.

PR #597, opened 2024-01-10
fix: Revert the clearCache change for API 33+

mykola-mokhnach:
Presumably fixes https://github.com/appium/appium-uiautomator2-server/issues/596

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
+++ b/app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
@@ -16,8 +16,6 @@
 
 package io.appium.uiautomator2.utils;
 
-import android.accessibilityservice.AccessibilityService;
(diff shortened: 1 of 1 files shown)


KazuCocoa (approved):
I'm good with this if this works over Android 13, but one thing in https://github.com/appium/appium/issues/19638 occurred in Android 12, which was `UiAutomatorBridge.getInstance().getUiAutomation().setServiceInfo(null);`. So maybe this won't help for Android 12

I can test this later with Android 13 (not soon but by tonight)

KazuCocoa:
Btw, my previous find an element and click/send keys to the element a couple of times did not cause the issue, so possibly @eglitise 's help would be needed to test with this.

mykola-mokhnach:
At least in the framework itself I observe no changes regarding the legacy API, which gives me more confidence.

PR #602, opened 2024-01-17
ci: use macos 13 arm64

KazuCocoa:


KazuCocoa:
hm, the instance was very slow to start, still no instances have picked up...

PR #603, opened 2024-01-17
test: open activity directly and add ignore to reduce flaky tests on CI env

KazuCocoa:
It looks like some failure occurred in the middle of `setupDragDropView`, so before running actual test


KazuCocoa:
Ignored tests succeeded in my local but not on the CI as... maybe mostly by slow emulator. Once the m1 mac instance is stable (currently triggering the instance took much time. My another pr was not able to get the instance after 10 minutes)

Issue #608, opened 2024-01-30
screenshot failed in real android device


kevinycm:
using ```webelement.screenshot(image_file_path)```，exception occured：

```
[AndroidUiautomator2Driver@45b4 (e0bb43ad)] Matched '/element' to command name 'findElement'
[AndroidUiautomator2Driver@45b4 (e0bb43ad)] Proxying [POST /element] to [POST http://127.0.0.1:8201/session/f496ffc8-866b-4285-9f9c-a764b7cc9624/element] with body: {"strategy":"id","selector":"com.tencent.mm:id/o7h","context":"","multiple":false}
[AndroidUiautomator2Driver@45b4 (e0bb43ad)] Got response with status 200: {"sessionId":"f496ffc8-866b-4285-9f9c-a764b7cc9624","value":{"ELEMENT":"00000000-0000-01fe-ffff-ffff000032d4","element-6066-11e4-a52e-4f735466cecf":"00000000-0000-01fe-ffff-ffff000032d4"}}
[AndroidUiautomator2Driver@45b4 (e0bb43ad)] Responding to client with driver.findElement() result: {"element-6066-11e4-a52e-4f735466cecf":"00000000-0000-01fe-ffff-ffff000032d4","ELEMENT":"00000000-0000-01fe-ffff-ffff000032d4"}
[HTTP] <-- POST /session/e0bb43ad-1604-4fdc-a241-35aad7fdb0bb/element 200 354 ms - 137
[HTTP]
[HTTP] --> GET /session/e0bb43ad-1604-4fdc-a241-35aad7fdb0bb/element/00000000-0000-01fe-ffff-ffff000032d4/screenshot
[HTTP] {}
[AndroidUiautomator2Driver@45b4 (e0bb43ad)] Calling AppiumDriver.getElementScreenshot() with args: ["00000000-0000-01fe-ffff-ffff000032d4","e0bb43ad-1604-4fdc-a241-35aad7fdb0bb"]
[AndroidUiautomator2Driver@45b4 (e0bb43ad)] Matched '/element/00000000-0000-01fe-ffff-ffff000032d4/screenshot' to command name 'getElementScreenshot'
[AndroidUiautomator2Driver@45b4 (e0bb43ad)] Did not know how to rewrite the original URL '/element/00000000-0000-01fe-ffff-ffff000032d4/screenshot' for W3C protocol
[...]

PR #609, opened 2024-01-30
fix: Recycling of screenshot bitmaps

mykola-mokhnach:
Addresses https://github.com/appium/appium-uiautomator2-server/issues/608

I turns out the `Bitmap.createBitmap` does not always create a new Bitmap instance and may also return the same instance that has been passed in arguments.


KazuCocoa (approved):
nice to find

PR #676, opened 2024-12-31
fix: stuck MJPEG clients when one client fails to initialize

vbarbaresi:
### Problem

There was an issue with the Client implementation: if a socket started a connection on the internal MJPEG port and was never initialized, the entire server thread would be stuck in the while loop waiting for the client to send something.

This broke other clients connected to the server: no frames would be received anymore.

To reproduce on emulator / real device:
- Start an Appium server with mJPEG server enabled (`appium:mjpegServerPort:9100` for instance)
- Start an Appium session that reads the mJPEG frames (using Appium Inspector for instance)
- Then from `adb shell`: connect to the internal mJPEG server (default port 7810): `nc localhost 7810`
- --> Streaming will be broken for the existing clients
- Another way to reproduce instead of using the phone shell would be from the host using the forwarded MJPEG port: `nc localhost 9100`

This happens because the client connection is not sending a proper HTTP request: it simply tries to receive data and this wasn't supported.
In this particular case we should not wait forever and not stop streaming frames to other clients

### Fix details

This patch fixes the problem by returning early: if the client is sending anything, initialization will be retried later if necessary instead of blocking the thread.

Another side effect is that we now actually receive frames using`nc localhost 9100`: this command was previously stuck waiting for the client to send something.
This can be useful for checking the status of the MJPEG server if we don't have a proper HTTP client handy

I added a unit test reproducing the bug by opening a socket connection to the MJPEG server and closing it without sending any data. Then verified that the server still works for another HTTP client. This test wouldn't pass before the fix

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -63,6 +63,8 @@ android {
                         '--add-exports', 'java.base/sun.nio.ch=ALL-UNNAMED',
                         '--add-opens', 'java.base/java.lang.reflect=ALL-UNNAMED',
                         '--add-opens', 'java.base/java.io=ALL-UNNAMED',
+                        '--add-opens', 'java.base/java.net=ALL-UNNAMED',
+                        '--add-opens', 'java.base/sun.net.www.protocol.http=ALL-UNNAMED',
                         '--add-exports', 'jdk.unsupported/sun.misc=ALL-UNNAMED',
                 ])
             }
--- a/app/src/test/java/io/appium/uiautomator2/server/MjpegScreenshotTest.java
+++ b/app/src/test/java/io/appium/uiautomator2/server/MjpegScreenshotTest.java
@@ -1,0 +1,75 @@
+package io.appium.uiautomator2.server.mjpeg;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.After;
+import org.junit.Before;
+import static org.junit.Assert.*;
+
+import java.net.URL;
+import java.net.HttpURLConnection;
+import java.util.Collections;
+import java.nio.charset.StandardCharsets;
+import java.net.Socket;
+
+import static org.mockito.Mockito.spy;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import io.appium.uiautomator2.server.mjpeg.MjpegScreenshotServer;
+import io.appium.uiautomator2.server.mjpeg.MjpegScreenshotStream;
+import io.appium.uiautomator2.server.ServerConfig;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(MjpegScreenshotStream.class)
+public class MjpegScreenshotTest {
+  private static MjpegScreenshotServer serverThread;
+  private static int streamingPort;
+
+  @Before
+  public void setUp() throws Exception {
+    // Create a MJPEG server with a mocked getScreenshot method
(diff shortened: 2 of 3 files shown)


app/src/test/java/io/appium/uiautomator2/server/MjpegScreenshotTest.java
```diff
@@ -0,0 +1,75 @@
+package io.appium.uiautomator2.server.mjpeg;
```
KazuCocoa, line 1: 👍

mykola-mokhnach (approved):

app/build.gradle
```diff
@@ -63,6 +63,8 @@ android {
                         '--add-exports', 'java.base/sun.nio.ch=ALL-UNNAMED',
                         '--add-opens', 'java.base/java.lang.reflect=ALL-UNNAMED',
                         '--add-opens', 'java.base/java.io=ALL-UNNAMED',
+                        '--add-opens', 'java.base/java.net=ALL-UNNAMED',
```
mykola-mokhnach, line 66: is the change needed for this PR?

vbarbaresi: yes, tests do not run for me without it:

```
MjpegScreenshotTest > shouldReceiveDataFromMjpegServer FAILED
    java.lang.reflect.InaccessibleObjectException: Unable to make static java.lang.String java.net.URL.toLowerCase(java.lang.String) accessible: module java.base does not "opens java.net" to unnamed module @79e4c792
        at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354)
        at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
        at java.base/java.lang.reflect.Method.checkCanSetAccessible(Method.java:199)
        at java.base/java.lang.reflect.Method.setAccessible(Method.java:193)
        at org.powermock.reflect.internal.WhiteboxImpl.doGetAllMethods(WhiteboxImpl.java:1508)
        at org.powermock.reflect.internal.WhiteboxImpl.getAllMethods(WhiteboxImpl.java:1482)
        at org.powermock.reflect.internal.WhiteboxImpl.getMethods(WhiteboxImpl.java:1750)
        at org.powermock.reflect.internal.WhiteboxImpl.getMethods(WhiteboxImpl.java:1789)
        at org.powermock.reflect.internal.WhiteboxImpl.getBestMethodCandidate(WhiteboxImpl.java:1008)
        at org.powermock.core.MockInvocation.findMethodToInvoke(MockInvocation.java:58)
        at org.powermock.core.MockInvocation.init(MockInvocation.java:35)
        at org.powermock.core.MockInvocation.<init>(MockInvocation.java:22)
        at org.powermock.core.MockGateway.doMethodCall(MockGateway.java:155)
        at org.powermock.core.MockGateway.methodCall(MockGateway.java:138)
        at io.appium.uiautomator2.server.mjpeg.MjpegScreenshotTest.shouldReceiveDataFromMjpegServer(MjpegScreenshotTest.java:51)
```

jlipps:
Hi @vbarbaresi, congrats: the Appium project wants to compensate you for this contribution! Please reply to this comment mentioning @jlipps and @KazuCocoa and sharing your [OpenCollective](https://opencollective.com/) account name, so that we can initiate payment! Or let us know if you decline to receive compensation via OpenCollective. Thank you!

vbarbaresi:
Hi, thanks for the proposition, it's much appreciated!
Unfortunately the contribution was made on behalf of the company so I have to decline the compensation.
I'm glad I contributed and I hope to be able to contribute again in the future!

jlipps:
Thanks for letting us know, hope to see you again @vbarbaresi :-)

PR #678, opened 2025-01-02
test: fix unit test flakiness on MjpegScreenshotTest

vbarbaresi:
I realized a bit too late that the tests I added in https://github.com/appium/appium-uiautomator2-server/pull/676 were flaky when  re-running the entire test suite:
` ./gradlew testServerDebugUnitTest   --rerun-tasks`

One of the tests would randomly (roughly 50% of the runs) fail with:
```
MjpegScreenshotTest > shouldNotBlockOnUnInitializedClient FAILED
    java.lang.AssertionError: expected:<200> but was:<-1>
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at org.junit.Assert.assertEquals(Assert.java:633)
        at io.appium.uiautomator2.server.mjpeg.MjpegScreenshotTest.shouldNotBlockOnUnInitializedClient(MjpegScreenshotTest.java:101)
```

I realized I was not sending a proper HTTP response, just some raw data on the socket (the HTTP header is added in getScreenshot() method, which was mocked)
As we send this data as streaming: the HTTP client doesn't really know how to interpret this data.
It was surprising that it worked in most of the test runs.

There is no need to test that we can really read this streaming response: just tweaking the mock data to send the proper header was enough to make the test reliable and see the response as 200.

- Skipped those tests in the CI using an environment variable to avoid flakiness issues
For the record: adding a `Thread.sleep(1000)` after starting the serverThread made the test pass in the CI, but it's not going to be 100% reliable. Easier to skip if for now

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -31,6 +31,8 @@ jobs:
       name: Run js linter
 
   java_test:
+    env:
+      CI: true
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
--- a/app/src/test/java/io/appium/uiautomator2/server/MjpegScreenshotTest.java
+++ b/app/src/test/java/io/appium/uiautomator2/server/MjpegScreenshotTest.java
@@ -12,6 +12,7 @@ import java.util.Collections;
 import java.nio.charset.StandardCharsets;
 import java.net.Socket;
 
+import org.junit.Assume;
 import static org.mockito.Mockito.spy;
 import org.powermock.api.mockito.PowerMockito;
 import org.powermock.core.classloader.annotations.PrepareForTest;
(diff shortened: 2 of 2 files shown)


mykola-mokhnach:
the PR prefix should rather be `test:`. It should not bump the module version

vbarbaresi:
Changed the PR prefix
Though it seems CI has failed 🤔 It was 100% reliable on my machine this time

mykola-mokhnach:
> Changed the PR prefix Though it seems CI has failed 🤔 It was 100% reliable on my machine this time

GH CI is slow, so it might (and usually does) affect the result.
You might skip the test conditionally if it is running in the CI env by setting the `CI` environment variable in the [workflow](https://github.com/appium/appium-uiautomator2-server/blob/master/.github/workflows/unit-test.yml) and then checking for its presence in the test.

vbarbaresi:
> You might skip the test conditionally if it is running in the CI env by setting the CI environment variable in the [workflow](https://github.com/appium/appium-uiautomator2-server/blob/master/.github/workflows/unit-test.yml) and then checking for its presence in the test.

Done :+1:
I also updated the PR description: test is now conditionally skipped in the CI.

mykola-mokhnach (approved):

PR #682, opened 2025-01-15
feat: add more fields from AccessibilityNodeInfo

ankahanets:
Add more field that the appium user can retrieve with getPageSource() or write tests that could previously only be done in the Espresso SDK, but not in the Appium SDK.


ankahanets:
Add more fields so that the appium user can retrieve with getPageSource() or write checks in their tests that could previously only be done in the Espresso SDK, but not in the Appium SDK.

app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
```diff
@@ -163,6 +168,49 @@ private static void putAttribute(Map<Attribute, Object> attribs, Attribute key,
...
+            case SCREEN_READER_FOCUSABLE:
+                return node.isScreenReaderFocusable();
+            case INPUT_TYPE:
+                return node.getInputType();
+            case DRAWING_ORDER:
+                return node.getDrawingOrder();
+            case SHOWING_HINT_TEXT:
+                return node.isShowingHintText();
+            case ACTIONS:
+//                return node.getActions(); //deprecated bitmask
```
mykola-mokhnach, line 182: Please remove the comment if it's not used

mykola-mokhnach:
Consider testing the performance of the updated solution. As you could [notice](https://github.com/appium/appium-uiautomator2-server/pull/682/files#diff-da49c51e46fb60a7096fe326dcfb53db0b03995b6607d136f4cac391e8bff84eR68) in the code we exclude "expensive" properties from the XML output as the attribute must be fetched for each node, which may significantly slow down the overall retrieval performance.

After the PR is merged please do not forget to document additional properties in https://github.com/appium/appium-uiautomator2-driver?tab=readme-ov-file#element-attributes

ankahanets:
Thanks for your input!

Added API level checks;

About performance, in the PR we add getters of already existing fields so it shouldn't significantly affect execution time. I can mention that XML parsing can be slower as each node has more attributes and to collect action list to one string we create a StringBuilder every time.

With my quick tests execution time looks a bit chaotic (with a debug build using `./gradlew app:installServerDebug`), but I get the next metrics.

```
getPageSource master
Screen #1: 543 659 270 873 1373 avg 743 w/o minmax 691
Screen #2: 771 214 1248 692 704 avg 725 w/o minmax 722
Screen #3: 573 555 1098 561 1094 avg 776 w/o minmax 742
Screen #4: 466 778 426 779 772 avg 644 w/o minmax 672

getPageSource PR
Screen #1: 864 1218 921 1254 762 avg 1003 w/o minmax 1001
Screen #2: 1024 788 1146 773 1068 avg 959 w/o minmax 960
Screen #3: 583 587 1118 578 1132 avg 799 w/o minmax 762
Screen #4: 812 800 467 451 446 avg 595 w/o minmax 572
```

I can investigate it more deeply. For example, I can try using deprecated `node.getActions()` instead of `node.getActionList()` and parse it's bitmask on the client side.

KazuCocoa:
Thank you for the performance measurements. This change could affect various actions (I guess)

One idea is to defend settings.
https://github.com/appium/appium-uiautomator2-driver?tab=readme-ov-file#settings-api
`Settings.get(SnapshotMaxDepth.class).getValue()` in the `UiElementSnapshot.java` could be a hint to implement that. So, for example... adding `pageSourceAdditionalAttributes` and the value is `all` or comma separate attribute names like `importantForAccessibility,textEntryKey` etc....

Just an idea, but like this thing also could be available. (The basic concept here is from  `pageSourceExcludedAttributes` in xcuitest driver)

app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
```diff
@@ -163,6 +168,56 @@ private static void putAttribute(Map<Attribute, Object> attribs, Attribute key,
...
+                    return null;
+                }
+            case INPUT_TYPE:
+                return node.getInputType();
+            case DRAWING_ORDER:
+                return node.getDrawingOrder();
+            case SHOWING_HINT_TEXT:
+                return node.isShowingHintText();
+            case ACTIONS:
+                StringBuilder actionsBuilder = new StringBuilder();
```
mykola-mokhnach, line 186: you might use TextUtils.join here

app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
```diff
@@ -163,6 +168,56 @@ private static void putAttribute(Map<Attribute, Object> attribs, Attribute key,
...
+            case LIVE_REGION:
+                return node.getLiveRegion();
+            case CONTEXT_CLICKABLE:
+                return node.isContextClickable();
+            case MAX_TEXT_LENGTH:
+                return node.getMaxTextLength();
+            case CONTENT_INVALID:
+                return node.isContentInvalid();
+            case ERROR_TEXT:
+                return node.getError();
```
mykola-mokhnach, line 218: please apply charSequenceToNullableString to all CharSequence types

app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
```diff
@@ -163,6 +168,56 @@ private static void putAttribute(Map<Attribute, Object> attribs, Attribute key,
                 } else {
                     return null;
                 }
+            case IMPORTANT_FOR_ACCESSIBILITY:
+                return node.isImportantForAccessibility();
+            case SCREEN_READER_FOCUSABLE:
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
```
mykola-mokhnach, line 174: you may use a ternary operator here to reduce the amount of boilerplate code

app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
```diff
@@ -163,6 +168,56 @@ private static void putAttribute(Map<Attribute, Object> attribs, Attribute key,
...
+                List<AccessibilityNodeInfo.AccessibilityAction> actionList = node.getActionList();
+                for (AccessibilityNodeInfo.AccessibilityAction action : actionList) {
+                    actionsBuilder.append(action.toString().split(" ")[1]);
+                    actionsBuilder.append(",");
+                }
+                if(actionsBuilder.length() > 0)
+                    actionsBuilder.deleteCharAt(actionsBuilder.length()-1);
+                return actionsBuilder.toString();
+            case TEXT_ENTRY_KEY:
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
```
mykola-mokhnach, line 196: same as above

app/src/main/java/io/appium/uiautomator2/utils/Attribute.java
```diff
@@ -45,6 +45,24 @@ public enum Attribute {
     TEXT(new String[]{"text", "name"}),
     HINT(new String[]{"hint"}),
     EXTRAS(new String[]{"extras"}),
+    IMPORTANT_FOR_ACCESSIBILITY(new String[]{"importantForAccessibility"}),
```
mykola-mokhnach, line 48: Long attribute names significantly increase resulting xml document size. Consider using shorter names, for example `axImportant`

app/src/main/java/io/appium/uiautomator2/utils/Attribute.java
```diff
@@ -45,6 +45,24 @@ public enum Attribute {
...
+    SCREEN_READER_FOCUSABLE(new String[]{"screenReaderFocusable"}),
+    INPUT_TYPE(new String[]{"inputType"}),
+    DRAWING_ORDER(new String[]{"drawingOrder"}),
+    SHOWING_HINT_TEXT(new String[]{"showingHintText"}),
+    TEXT_ENTRY_KEY(new String[]{"textEntryKey"}),
+    MULTI_LINE(new String[]{"multiLine"}),
+    DISMISSABLE(new String[]{"dismissable"}),
+    ACCESSIBILITY_FOCUSED(new String[]{"accessibilityFocused"}),
```
mykola-mokhnach, line 56: axFocused

app/src/main/java/io/appium/uiautomator2/utils/Attribute.java
mykola-mokhnach: also, as you could see above we used to split words with dashes: `selection-end` vs `selectionEnd`. I'd suggest having the same style in below attribute names

app/src/main/java/io/appium/uiautomator2/utils/Attribute.java
```diff
@@ -45,6 +45,24 @@ public enum Attribute {
...
+    HEADING(new String[]{"heading"}),
+    LIVE_REGION(new String[]{"liveRegion"}),
+    CONTEXT_CLICKABLE(new String[]{"contextClickable"}),
+    MAX_TEXT_LENGTH(new String[]{"maxTextLength"}),
+    CONTENT_INVALID(new String[]{"contentInvalid"}),
+    ERROR_TEXT(new String[]{"errorText", "error"}),
+    PANE_TITLE(new String[]{"paneTitle"}),
+    ACTIONS(new String[]{"actions"}),
+
```
mykola-mokhnach, line 65: it makes sense to define new items after CONTENT_SIZE to preserve the ordinality

mykola-mokhnach:
@KazuCocoa I've checked the actual source code for the newly added attributes and they seem to be simple getters, so I don't expect much influence to the actual performance. 

Although, I tend to agree to the point that maybe not of all these attributes are actually useful in the XML tree output by default. 

@ankahanets What exactly do you want to achieve by adding all of these attributes to the page source? Are they equally useful for your automation scenarios? Perhaps, we could still have all of them available for attributes fetching, but only have some (most useful ones) in the actual XML document?

app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
ankahanets: Unfortunately, `TextUtils.join(",", node.getActionList());` will not work here, as `AccessibilityAction::toString` provide something of this: `AccessibilityAction: ACTION_SELECT - null`. So I think to leave it with StringBuilder.

app/src/main/java/io/appium/uiautomator2/utils/Attribute.java
ankahanets: Will use `a11y-important` as more common abbreviation

app/src/main/java/io/appium/uiautomator2/utils/Attribute.java
ankahanets: "a11y-focused"

ankahanets:
@mykola-mokhnach We are developing an Appium/Espresso tool for checking accessibility correctness, and we need some of the data for our internal validations.

I can remove the PANE_TITLE, ERROR_TEXT, and CONTENT_INVALID fields since we don’t use them right now. I added them “just in case for the future” and to make the input more similar to what we have in Espresso, but other fields will be really helpful for our SDK.

However, if the number of new fields is too large, we can introduce a new settings option and return the most obscure field for the ordinal user only if it is turned on in settings (like we have for extras). But I think that most of the new fields will be useful for different users as well.

Also, to reduce the xml size we can return nulls for the default values, like I've done for `MAX_TEXT_LENGTH`. The default value is -1, so there will be no such attribute for non-EditText nodes.

app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
```diff
@@ -158,11 +164,49 @@ private static void putAttribute(Map<Attribute, Object> attribs, Attribute key,
...
-                    return node.getHintText();
-                } else {
-                    return null;
+                return Build.VERSION.SDK_INT >= Build.VERSION_CODES.O ? node.getHintText() : null;
+            case IMPORTANT_FOR_ACCESSIBILITY:
+                return Build.VERSION.SDK_INT >= Build.VERSION_CODES.N ? node.isImportantForAccessibility() : null;
+            case SCREEN_READER_FOCUSABLE:
+                return Build.VERSION.SDK_INT >= Build.VERSION_CODES.P ? node.isScreenReaderFocusable() : null;
+            case INPUT_TYPE:
+                return node.getInputType() != 0 ? node.getInputType() : null;
```
mykola-mokhnach, line 173: I assume having zero here is ok

app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
```diff
@@ -158,11 +164,49 @@ private static void putAttribute(Map<Attribute, Object> attribs, Attribute key,
...
+            case DRAWING_ORDER:
+                return Build.VERSION.SDK_INT >= Build.VERSION_CODES.N ? node.getDrawingOrder() : null;
+            case SHOWING_HINT_TEXT:
+                return Build.VERSION.SDK_INT >= Build.VERSION_CODES.O ? node.isShowingHintText() : null;
+            case ACTIONS:
+                StringBuilder actionsBuilder = new StringBuilder();
+                List<AccessibilityNodeInfo.AccessibilityAction> actionList = node.getActionList();
+                for (AccessibilityNodeInfo.AccessibilityAction action : actionList) {
+                    actionsBuilder.append(action.toString().split(" ")[1]);
```
mykola-mokhnach, line 182: to avoid unexpected exceptions we need to make sure that action is not null and that the result of split has at least 2 items

app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
```diff
@@ -158,11 +164,49 @@ private static void putAttribute(Map<Attribute, Object> attribs, Attribute key,
...
+                    actionsBuilder.append(",");
                 }
+                if(actionsBuilder.length() > 0)
```
mykola-mokhnach, line 185: Please always use curly brackets for condition blocks, even if they only have one expression

app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
```diff
@@ -158,11 +164,49 @@ private static void putAttribute(Map<Attribute, Object> attribs, Attribute key,
...
+                    actionsBuilder.deleteCharAt(actionsBuilder.length()-1);
```
mykola-mokhnach, line 186: Please format the expression

app/src/main/java/io/appium/uiautomator2/utils/Attribute.java
```diff
@@ -45,13 +45,32 @@ public enum Attribute {
...
     // it does not replace null values with empty strings
     ORIGINAL_TEXT(new String[]{"original-text"}, false, false),
     BOUNDS(new String[]{"bounds"}),
     INDEX(new String[]{"index"}, false, true),
     DISPLAYED(new String[]{"displayed"}),
-    CONTENT_SIZE(new String[]{"contentSize"}, true, false);
+    CONTENT_SIZE(new String[]{"contentSize"}, true, false),
+
+    IMPORTANT_FOR_ACCESSIBILITY(new String[]{"a11y-important", "importantForAccessibility"}),
+    SCREEN_READER_FOCUSABLE(new String[]{"screenReaderFocusable"}),
```
mykola-mokhnach, line 58: please apply the common naming style to this attribute as well. Also, maybe there is a possibility to make the name shorter

app/src/main/java/io/appium/uiautomator2/utils/Attribute.java
```diff
@@ -45,13 +45,32 @@ public enum Attribute {
...
+    DISMISSABLE(new String[]{"dismissable"}),
+    ACCESSIBILITY_FOCUSED(new String[]{"a11y-focused", "accessibilityFocused"}),
+    HEADING(new String[]{"heading"}),
+    LIVE_REGION(new String[]{"live-region", "liveRegion"}),
+    CONTEXT_CLICKABLE(new String[]{"context-clickable", "contextClickable"}),
+    MAX_TEXT_LENGTH(new String[]{"max-text-length", "maxTextLength"}),
+    CONTENT_INVALID(new String[]{"content-invalid", "contentInvalid"}),
+    ERROR_TEXT(new String[]{"error", "errorText"}),
+    PANE_TITLE(new String[]{"pane-title", "paneTitle"}),
+    ACTIONS(new String[]{"actions"});
```
mykola-mokhnach, line 73: I would say we should not include actions to the xml by default as the resulting string might be pretty long. Consider putting under a setting

app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
```diff
@@ -158,11 +164,49 @@ private static void putAttribute(Map<Attribute, Object> attribs, Attribute key,
...
+            case SCREEN_READER_FOCUSABLE:
+                return Build.VERSION.SDK_INT >= Build.VERSION_CODES.P ? node.isScreenReaderFocusable() : null;
+            case INPUT_TYPE:
+                return node.getInputType() != 0 ? node.getInputType() : null;
+            case DRAWING_ORDER:
+                return Build.VERSION.SDK_INT >= Build.VERSION_CODES.N ? node.getDrawingOrder() : null;
+            case SHOWING_HINT_TEXT:
+                return Build.VERSION.SDK_INT >= Build.VERSION_CODES.O ? node.isShowingHintText() : null;
+            case ACTIONS:
+                StringBuilder actionsBuilder = new StringBuilder();
```
mykola-mokhnach, line 179: consider moving this block of code (compiling actions to an attribute string) to a separate helper method

app/src/main/java/io/appium/uiautomator2/utils/Attribute.java
ankahanets: Missed this, fixed.
But didn't find how to make it shorter and keep clear.

app/src/main/java/io/appium/uiautomator2/utils/Attribute.java
ankahanets: Hi, I added new settings field `includeA11yActionsInPageSource` for this case, similar to the `includeExtrasInPageSource`, please take a look.

app/src/main/java/io/appium/uiautomator2/model/BaseElement.java
```diff
@@ -269,4 +270,25 @@ public static String getExtrasAsString(AccessibilityNodeInfo nodeInfo) {
...
+    @Nullable
+    public static String getA11yActionsAsString(AccessibilityNodeInfo nodeInfo) {
+        StringBuilder actionsBuilder = new StringBuilder();
+        List<AccessibilityAction> actionList = nodeInfo.getActionList();
+        if (actionList.isEmpty()) {
+            return null;
+        }
+        for (AccessibilityAction action : actionList) {
+            String[] split = action.toString().split(" ");
+            if (split.length < 2) {
```
mykola-mokhnach, line 283: Cannot we just return a single action in such case?

app/src/main/java/io/appium/uiautomator2/model/BaseElement.java
```diff
@@ -269,4 +270,25 @@ public static String getExtrasAsString(AccessibilityNodeInfo nodeInfo) {
...
+                return null;
```
mykola-mokhnach, line 284: should we continue instead?

app/src/main/java/io/appium/uiautomator2/model/BaseElement.java
```diff
@@ -269,4 +270,25 @@ public static String getExtrasAsString(AccessibilityNodeInfo nodeInfo) {
...
+            }
+            actionsBuilder.append(split[1]);
```
mykola-mokhnach, line 286: appeand returns this, so you could chain these calls

app/src/main/java/io/appium/uiautomator2/model/settings/IncludeA11yActionsInPageSource.java
```diff
@@ -0,0 +1,42 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.uiautomator2.model.settings;
+
+public class IncludeA11yActionsInPageSource extends AbstractSetting<Boolean> {
+    private static final String SETTING_NAME = "includeA11yActionsInPageSource";
```
mykola-mokhnach, line 20: Please also do not forget this new setting into settings documentation after this PR is released

app/src/main/java/io/appium/uiautomator2/model/BaseElement.java
```diff
@@ -269,4 +270,25 @@ public static String getExtrasAsString(AccessibilityNodeInfo nodeInfo) {
         }
         return TextUtils.join(EXTRAS_SEPARATOR, extras);
     }
+
+    @Nullable
+    public static String getA11yActionsAsString(AccessibilityNodeInfo nodeInfo) {
+        StringBuilder actionsBuilder = new StringBuilder();
```
mykola-mokhnach, line 276: the builder instance should be constructed after the guard

app/src/main/java/io/appium/uiautomator2/model/BaseElement.java
```diff
@@ -269,4 +270,25 @@ public static String getExtrasAsString(AccessibilityNodeInfo nodeInfo) {
...
+        List<AccessibilityAction> actionList = nodeInfo.getActionList();
+        if (actionList.isEmpty()) {
+            return null;
+        }
+        for (AccessibilityAction action : actionList) {
+            String[] split = action.toString().split(" ");
```
mykola-mokhnach, line 282: it would be useful for readability to add a comment with an example about how a raw value looks like before we split it

app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
```diff
@@ -17,6 +17,7 @@
 package io.appium.uiautomator2.model;
 
 import android.os.Build;
+import android.text.TextUtils;
```
mykola-mokhnach, line 20: is this import used?

app/src/main/java/io/appium/uiautomator2/model/BaseElement.java
ankahanets: No... actually is not possible, but even if this happens - than means something strange and all the data is broken

app/src/main/java/io/appium/uiautomator2/model/BaseElement.java
ankahanets: yes... I think it looks more clear now, but I can change if it's needed indeed

mykola-mokhnach (approved):

mykola-mokhnach:
@KazuCocoa Do you have any more comments to this PR?

KazuCocoa (approved):
No, lgtm 🚀

jlipps:
Hi @ankahanets, congrats: the Appium project wants to compensate you for this contribution! Please reply to this comment mentioning @jlipps and @KazuCocoa and sharing your [OpenCollective](https://opencollective.com/) account name, so that we can initiate payment! Or let us know if you decline to receive compensation via OpenCollective. Thank you!

ankahanets:
> Hi @ankahanets, congrats: the Appium project wants to compensate you for this contribution! Please reply to this comment mentioning @jlipps and @KazuCocoa and sharing your [OpenCollective](https://opencollective.com/) account name, so that we can initiate payment! Or let us know if you decline to receive compensation via OpenCollective. Thank you!

@jlipps sorry, only today noticed your comment, I will check what is OpenCollective and answer about account a bit later.

ankahanets:
@jlipps Hi, created an account, hope that's what we need
https://opencollective.com/andrii-kahanets

KazuCocoa:
Thank you! We'll treat the opencollective stuff tonight

PR #690, opened 2025-03-18
feat: Replace the deprecated AndroidJUnit4 class in the runner

mykola-mokhnach:
Also did some refactoring to the existing code to streamline the InstrunemntationServer behaviour


mykola-mokhnach:
@KazuCocoa Could you please help to test this change with UIA2 driver before we publish it?

app/build.gradle
```diff
@@ -102,6 +102,7 @@ dependencies {
     implementation 'com.github.vbmacher:java-cup-runtime:11b-20160615-3'
     //noinspection GradleDependency
     implementation 'com.ibm.icu:icu4j:60.2'
+    implementation 'androidx.test.ext:junit:1.2.1'
```
KazuCocoa, line 105: I think if we added this in `implementation`, others can be removed since they testImplementation/androidTestImplementation inherit the `implementation`

KazuCocoa (commented):
(not tested with real server yet)

app/build.gradle
mykola-mokhnach: removed

KazuCocoa (approved):
Left one more deps thing, otherwise lg

PR #698, opened 2025-04-17
feat(build): migrate Gradle build files from Groovy DSL (build.gradle) to Kotlin DSL (build.gradle.kts)

edsion1107:
- Add BUILD_TIME to BuildConfig and print at server start
- Migrated to Kotlin DSL: Gradle build files (build.gradle, settings.gradle) have been converted from Groovy DSL to Kotlin DSL (build.gradle.kts, settings.gradle.kts).
- Gradle Version Catalog: Dependency management is now centralized using Gradle Version Catalog (gradle/libs.versions.toml). Direct dependency declarations in build.gradle.kts are replaced with references to the catalog.

BREAKING CHANGE: Migrated to Kotlin DSL & Gradle Version Catalog


mykola-mokhnach:
Thank you @edsion1107 
Could you please take a look at failing CI e2e tests?

app/src/androidTestServer/java/io/appium/uiautomator2/server/test/AppiumUiAutomator2Server.java
```diff
@@ -35,11 +43,47 @@ public class AppiumUiAutomator2Server {
...
+     * <p>This method uses reflection to collect static field values from the BuildConfig class
+     * and converts them into a JSON format using Gson library.</p>
+     *
+     * @return JSON string representation of BuildConfig's static fields
+     * @throws RuntimeException if BuildConfig class cannot be found or field access is denied
+     *
+     * @apiNote The BuildConfig class name is hardcoded to "io.appium.uiautomator2.test.BuildConfig".
+     *          Modify the class name if used in different packages.
+     */
+    public static String buildConfigSerializer() {
```
mykola-mokhnach, line 66: method names should contain verbs. How about renaming this one to `serializeBuildConfig()` ?

app/src/androidTestServer/java/io/appium/uiautomator2/server/test/AppiumUiAutomator2Server.java
```diff
@@ -35,11 +43,47 @@ public class AppiumUiAutomator2Server {
...
+        final String buildConfigClassName = "io.appium.uiautomator2.test.BuildConfig";
```
mykola-mokhnach, line 67: should it be a static class constant?

mykola-mokhnach: I would also prefer to have it imported explicitly, like `BuildConfig.getName()`. Having this constant as a string might lead to potential issues if the original class is moved/renamed

app/src/androidTestServer/java/io/appium/uiautomator2/server/test/AppiumUiAutomator2Server.java
```diff
@@ -35,11 +43,47 @@ public class AppiumUiAutomator2Server {
...
+            Class<?> buildConfigClass = Class.forName(buildConfigClassName);
+
+            for (Field field : buildConfigClass.getDeclaredFields()) {
+                if (Modifier.isStatic(field.getModifiers())) {
+                    try {
+                        field.setAccessible(true);
+                        Object value = field.get(null);
+                        fieldMap.put(field.getName(), value);
+                    } catch (IllegalAccessException e) {
+                        throw new RuntimeException("Field access denied for: " + field.getName(), e);
```
mykola-mokhnach, line 80: I assume this method call should be safe as it is only used for logging. All exceptions could simply be dumped to logs, but it should not fail the server execution

app/src/androidTestServer/java/io/appium/uiautomator2/server/test/AppiumUiAutomator2Server.java
```diff
@@ -35,11 +43,47 @@ public class AppiumUiAutomator2Server {
     @Test
     public void startServer() throws InterruptedException {
         Logger.info( ">>>Entry point start<<<");
+        Logger.info(buildConfigSerializer());
```
mykola-mokhnach, line 46: could you please provide an example output of this method?

mykola-mokhnach:
Does the current [dependabot config](https://github.com/appium/appium-uiautomator2-server/blob/223e4a8a8d428fdd5269f266837bd5c90e55a198/.github/dependabot.yml#L12) support the new versioning scheme? Could you please verify that and add the necessary changes if it does not?

app/src/androidTestServer/java/io/appium/uiautomator2/server/test/AppiumUiAutomator2Server.java
edsion1107: I understand your point. When I moved `android.defaults.buildfeatures.buildconfig=true` from `gradle.properties` to `app/build.gradle.kts`, I added a custom `BUILD_TIME` field in `buildConfigFields` and tested it via the `buildConfigSerializer` method. However, there are indeed issues such as non-standard naming conventions and flaws in the implementation details. I plan to address these issues later.

app/src/androidTestServer/java/io/appium/uiautomator2/server/test/AppiumUiAutomator2Server.java
mykola-mokhnach: it would be nice to add this information into the /status API output as well, similarly to how it works for Espresso and WDA

edsion1107:
> Does the current [dependabot config](https://github.com/appium/appium-uiautomator2-server/blob/223e4a8a8d428fdd5269f266837bd5c90e55a198/.github/dependabot.yml#L12) support the new dependencies versioning scheme? Could you please verify that and add the necessary changes if it does not?

I noticed that the previous Gradle dependency tracking configuration was no longer functioning correctly, as the repository's Dependency Graph was only showing Node.js-related dependencies. Following the [documentation](https://github.com/marketplace/actions/build-with-gradle#the-dependency-submission-action), I implemented the GitHub Action-based dependency submission workflow for Gradle.

app/src/androidTestServer/java/io/appium/uiautomator2/server/test/AppiumUiAutomator2Server.java
edsion1107: Thanks for the suggestion. I maintained BuildConfig logging in logcat for debugging. Regarding API changes:

1. Compatibility: Modifying StatusModel might cause deserialization issues in Java/Python clients (needs verification).

2. Scope: Since this PR focuses on Gradle builds, should we handle API changes in a separate PR?

This keeps build changes isolated and allows proper client testing.

mykola-mokhnach: I am totally fine with the plan to do that in a separate PR.

My point is that if we would like to integrate this functionality with the above endpoint then the static helper method which is currently privately isolated in the test entry point class could not be reused for such purpose.
Also, it is not very clear what exactly it returns in the serialized json string.

Based on that I assume that we anyway would like to refactor this serialized method sooner or later:
- move it to shared utilities class
- create a data model and map actual build config properties to this data model fields explicitly, so the method could return an explcit data
- Extract serialization itself from this method, so we could decide if the calling code if we want the model being serialized or not

It would also work for me if we keep only .kts migration as part of this PR and do the above stuff in a separate PR as well.

What do you think @edsion1107 ?

mykola-mokhnach:
@KazuCocoa Do you think this PR is breaking?
I personally don't see any major changes there as from a client perspective the build script migration to Kotlin should be transparent.

app/src/androidTestServer/java/io/appium/uiautomator2/server/test/AppiumUiAutomator2Server.java
edsion1107: I fully understand your points regarding the API implementation. To address the testing aspects:

1. Test Execution Conflicts:
When alternating between `connectedE2eTestDebugAndroidTest` and `connectedServerDebugAndroidTest` tasks locally, occasional "Address Already in Use" errors may occur. These can be traced back via logs by checking the active productFlavor and buildType.

2. Build Validation:
Since versionCode/versionName remain static during local development, I'm using BuildConfig.BUILD_TIME to:
- Verify installation of latest builds
- Validate Gradle incremental compilation
- Confirm build cache effectiveness

The reflection-based approach was chosen to:
- Maintain debug logging compatibility
- Future-proof against BuildConfig field additions

I recommend keeping the current minimal debug logging implementation. For future API changes, we should adopt shared utility classes, data models, and corresponding serialization approaches.

KazuCocoa:
I think this is a feat or chore level one without any breakings as converting build script from groovy to kotlin

scripts/gradle-version-update.js
```diff
@@ -4,8 +4,8 @@ const semver = require('semver');
 const {logger} = require('@appium/support');
 
 const log = logger.getLogger('Versioner');
-const VERSION_NAME_PATTERN = /^\s*versionName\s+['"](.+)['"]$/gm;
-const VERSION_CODE_PATTERN = /^\s*versionCode\s+(.+)$/gm;
+const VERSION_NAME_PATTERN = /^\s*versionName\s*=\s*(.+)$/gm;
```
KazuCocoa, line 7: 👍

package.json
```diff
@@ -28,7 +28,7 @@
   ],
   "homepage": "https://github.com/appium/appium-uiautomator2-server",
   "scripts": {
-    "bump-gradle-version": "node ./scripts/gradle-version-update.js --package-version=${npm_package_version} && git add app/build.gradle",
+    "bump-gradle-version": "node ./scripts/gradle-version-update.js --package-version=${npm_package_version} && git add app/build.gradle.kts",
```
KazuCocoa, line 31: Should this git add be `gradle.properties` to commit the version update by `gradle-version-update.js`?

app/build.gradle.kts
```diff
@@ -0,0 +1,241 @@
...
+// Apply UnMock plugin via legacy syntax because it's not properly published
+apply(plugin = "de.mobilej.unmock")
+plugins {
+    alias(libs.plugins.android.application)
+}
+
+java {
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(22)
+        vendor = JvmVendorSpec.ADOPTIUM
```
KazuCocoa, line 23: Do we need to specify this vendor?

app/src/androidTestServer/java/io/appium/uiautomator2/server/test/AppiumUiAutomator2Server.java
```diff
@@ -16,12 +16,20 @@
 package io.appium.uiautomator2.server.test;
 
 import androidx.test.ext.junit.runners.AndroidJUnit4;
-
+import com.google.gson.Gson;
```
KazuCocoa, line 19: Do we need diff in this file?

app/build.gradle.kts
edsion1107: image
When testing locally via the command line, I tried using the same JDK version bundled with Android Studio (jetbrains-jdk-21.0.6b631.42) but encountered the error mentioned above.

Additionally, I'm following the Java configuration patterns established in our functional-test workflows to:

1. Prevent breaking changes

2. Maintain consistency between CI and local environments

3. Avoid potential issues caused by environment discrepancies

This approach ensures our development pipeline remains stable while aligning with the team's existing practices.

KazuCocoa: Thanks. I currently use Android Studio's embedded Java env, so the jdk is 21. Then, 

```
java {
    toolchain {
        languageVersion = JavaLanguageVersion.of(21)
        vendor = JvmVendorSpec.ADOPTIUM
    }
}
```

had an error about:

```
A build operation failed.
    Could not create task ':app:compileE2eTestDebugAndroidTestJavaWithJavac'.
Could not create task ':app:compileE2eTestDebugAndroidTestJavaWithJavac'.
Failed to calculate the value of task ':app:compileE2eTestDebugAndroidTestJavaWithJavac' property 'javaCompiler'.
Cannot find a Java installation on your machine matching this tasks requirements: {languageVersion=21, vendor=Eclipse Temurin, implementation=vendor-specific} for MAC_OS on aarch64.
No locally installed toolchains match and toolchain auto-provisioning is not enabled.
```

`JvmVendorSpec.JETBRAINS` worked so I asked this. My local had no workable env for now with `JvmVendorSpec.ADOPTIUM` but CI is fine, so I guess this PR is ok.

KazuCocoa: btw, could you leave a comment about this in the lines for future maintenance, why they are specified?

app/src/androidTestServer/java/io/appium/uiautomator2/server/test/AppiumUiAutomator2Server.java
KazuCocoa: This is remaining. Please revert this file's diff if they are no longer needed

mykola-mokhnach (approved):

KazuCocoa (approved):

KazuCocoa:
Hi @edsion1107, congrats: the Appium project wants to compensate you for this contribution! Please reply to this comment mentioning @jlipps and @KazuCocoa and sharing your [OpenCollective](https://opencollective.com/) account name, so that we can initiate payment! Or let us know if you decline to receive compensation via OpenCollective. Thank you!

edsion1107:
Thank you for the opportunity and recognition! I'm honored to contribute to the Appium project. My OpenCollective account is edsion1107, and the associated email is zjb861107@gmail.com. Please let me know if you need any additional information to proceed with the payment. Thanks again!

cc: @jlipps @KazuCocoa

PR #705, opened 2025-05-31
ci: add api level 36

KazuCocoa:


KazuCocoa:
These test failures probably would make sense to check:

- toastVerificationTest1
- toastVerificationTest2
- toastVerificationTest3
- toastVerificationTest4

To the same toast's test https://github.com/appium/appium-uiautomator2-driver/blob/3965602f04cc6f5818b67f2d6bba6c6b0a9b6167/test/functional/commands/find/by-xpath-e2e-specs.js#L69-L87 passed though.

PR #711, opened 2025-07-22
 feat: add tooltip-text and text-has-clickable-span attributes to UI elements

ankahanets:
We previously worked on a similar PR that added more fields to the UI elements here:
https://github.com/appium/appium-uiautomator2-server/pull/682

Now, I’d like to add a few additional fields:

 - `tooltip-text` - The value of AccessibilityNodeInfo::getTooltipText() - In many cases, this data can be obtained from the `content-desc` field, but for some apps a separate field is necessary (particularly for Flutter and React Native apps);

 - `text-has-clickable-span` - A new custom field that describes whether there are ClickableSpan or URLSpan elements in the TextView content. We need to check if the TextView has clickable elements and thus actually can be focused.
 
 Both fields are empty for most elements and will appear only on specific nodes, so there will be no page source bloat.

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/UiElementSnapshot.java
@@ -17,6 +17,9 @@
 package io.appium.uiautomator2.model;
 
 import android.os.Build;
+import android.text.Spanned;
+import android.text.style.ClickableSpan;
+import android.text.style.URLSpan;
 import android.util.Pair;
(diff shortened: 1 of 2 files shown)


mykola-mokhnach (approved):
Please do not forget to document new properties after the change is published

ankahanets:
> Please do not forget to document new properties after the change is published

@mykola-mokhnach docs are here:
https://github.com/appium/appium-uiautomator2-driver/pull/912

PR #734, opened 2025-11-29
feat: calls clearCache from UiAutomation API for API level 34+

KazuCocoa:
I found out https://developer.android.com/reference/android/app/UiAutomation#clearCache() was available since api level 34.

- https://developer.android.com/reference/android/app/UiAutomation#clearCache()
- https://android.googlesource.com/platform/frameworks/base/+/refs/heads/android14-dev/core/java/android/app/UiAutomation.java#519

What about calling this for 34+?

I also marked this as `feat` since it may be better in the minor version change rather than a patch

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
+++ b/app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
@@ -49,7 +49,7 @@ public class AXWindowHelpers {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
                 boolean cleared = UiAutomatorBridge.getInstance().getUiAutomation().clearCache();
                 if (!cleared) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/appium/uiautomator2/utils/AXWindowHelpers.java
```diff
@@ -46,8 +46,15 @@ public class AXWindowHelpers {
      */
     private static void clearAccessibilityCache() {
         try {
-            // This call invokes `AccessibilityInteractionClient.getInstance().clearCache();` method
-            UiAutomatorBridge.getInstance().getUiAutomation().setServiceInfo(null);
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+                boolean cleared = UiAutomatorBridge.getInstance().getUiAutomation().clearCache();
```
mykola-mokhnach: should we put this call under try/except?

mykola-mokhnach (approved):

PR #735, opened 2025-11-29
chore: Respect CurrentDisplayId setting while taking screenshots

mykola-mokhnach:


app/src/main/java/io/appium/uiautomator2/utils/ScreenshotHelper.java
```diff
@@ -82,64 +87,118 @@ public static String takeScreenshot() throws TakeScreenshotException {
...
+
         UiAutomation automation = CustomUiDevice.getInstance().getUiAutomation();
-        DisplayMetrics metrics = new DisplayMetrics();
-        display.getMetrics(metrics);
-        Bitmap screenshot = null;
-        Logger.debug(String.format("Display metrics: %s", metrics));
-        // Workaround for https://github.com/appium/appium/issues/12199
-        // executeShellCommand seems to be faulty on Android 5
-        if (metrics.densityDpi != DENSITY_DEFAULT) {
+        if (shouldTryScreencap) {
```
KazuCocoa: I think we should keep https://github.com/appium/appium/issues/12199 comment to not forget the reason. A newly created default emulator also went to this `shouldTryScreencap == true` condition as the density was not `DENSITY_DEFAULT`(160). - I tested locally

Also, then, I think... we no longer need to have `Bitmap screenshot = automation.takeScreenshot();` (L127) way since the reason is Android 5 (according to the comment), which we stopped supporting. Then, we can remove `if (shouldTryScreencap) {` condition as well

PR #739, opened 2025-11-30
feat: Add API to list windows

mykola-mokhnach:


app/src/main/java/io/appium/uiautomator2/handler/ListWindows.java
```diff
@@ -0,0 +1,271 @@
...
+        }
+
+        try {
+            Window window = getWindowFromAccessibilityWindowInfo(windowInfo);
+            if (window == null) {
+                return null;
+            }
+
+            UiAutomation uiAutomation = CustomUiDevice.getInstance().getUiAutomation();
+            Bitmap screenshot = uiAutomation.takeScreenshot(window);
```
KazuCocoa: I see... They could have null cases depending on the condition

https://developer.android.com/reference/android/app/UiAutomation#takeScreenshot(android.view.Window)

> Used to capture a screenshot of a Window. This can return null in the following cases: 1. Window content hasn't been layed out. 2. Window doesn't have a valid SurfaceControl 3. An error occurred in SurfaceFlinger when trying to take the screenshot.

We could address this in the command's readme or log for `null` case (for future user support)

app/src/main/java/io/appium/uiautomator2/model/api/WindowModel.java
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.uiautomator2.model.api;
+
+import android.graphics.Rect;
+import androidx.annotation.Nullable;
+
+public class WindowModel {
```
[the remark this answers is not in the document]
mykola-mokhnach: lets do it in a separate PR, this one is already big enough

app/src/main/java/io/appium/uiautomator2/handler/ListWindows.java
mykola-mokhnach: yep, will mention it in the method readme

KazuCocoa (approved):

Issue #774, opened 2026-06-25
Server: unbounded dalvik-LinearAlloc growth under sustained findElement load (ART arena; not class-loading or JIT)


philip-hh:
> Orthogonal to the `AccessibilityNodeInfo` recycling fix (#770) — this is a separate, pre-existing leak.
>
> **Note:** not yet a reliable repro — the accumulation precondition is unidentified and the leaky window is ~2 runs off fresh state; expect to re-establish state between attempts.

## Summary

Under sustained, zero-think-time `findElement` load, the on-device UiAutomator2 server process (`io.appium.uiautomator2.server`) grows its **`[anon:dalvik-LinearAlloc]`** region without bound (observed ~1 GB in a single hammer run; large enough to be a real contributor to low-memory pressure on constrained devices). The growth is **ART arena allocation**, not application heap, and is **not** explained by class loading or JIT.

## Environment

- Android 12 (API 31/32), arm, real hardware.
- UiAutomator2 server built as the standard **debuggable** instrumentation APK (required to run under `am instrument`).
- Driven via appium-uiautomator2-driver; locator strategy in the repro is **accessibility id** (no XPath / no page-source path involved).

## Symptom

- `/proc/<pid>/smaps` shows **`[anon:dalvik-LinearAlloc]`** climbing roughly linearly with find volume during a leaky run (single hammer run observed growing into the hundreds of MB to ~1 GB).
- All other regions are flat: Java heap, native heap, parcel, binder buffers, message/event queues.

## Ruled out

- **Class loading.** `-verbose:class` during a live leaky hammer shows the process loads its classes at startup and then **zero** additional classes while LinearAlloc grows. So: no transient ClassLoaders, no dynamic proxies, no in-memory dex.
- **JIT.** Re-confirmed with JIT enabled (default): a run with JIT on stayed flat through thousands of finds. An earlier "flat with JIT disabled" result was a confound (the trigger had depleted), not a JIT effect.
- **Application-level objects.** Java/native heap and parcel/binder are flat; the growth is purely the ART LinearAlloc arena.
- **AccessibilityNodeInfo recycling.** Zero `IllegalStateException`/"already recycled"; the node lifecycle is clean (and the recycling fix in #770 is a separate concern).

## Characterization

- It is a **per-find ART arena (LinearAlloc) allocation in a *fresh* server process** that is **heavy-then-light depending on accumulated device-internal state**.
[...]

PR #775, opened 2026-06-28
feat: Add deferAccessibilityCacheReset setting to bound LinearAlloc growth under sustained finds

philip-hh:
Refs #774 — server-side mitigation. (The deeper ART conflict-table re-allocation question stays tracked in #774.)

## Problem
Clearing the `AccessibilityInteractionClient` cache before *every* find/source lookup (`AXWindowHelpers.resetAccessibilityCache()` → `clearCache()` / `setServiceInfo(null)`) forces a full Binder re-traversal of the accessibility tree each time. Under sustained zero-think-time finds this drives unbounded **ART `[anon:dalvik-LinearAlloc]`** growth — IMT conflict-table / method-resolution metadata (`art::ClassLinker::AddMethodToConflictTable` → `LinearAlloc::Alloc` → `MemMap::MapAnonymous` → `mmap`), attributed via simpleperf in #774 — which can OOM the server on long runs.

## Change
New **opt-in** setting **`deferAccessibilityCacheReset`** (default `false`). When enabled, the per-find cache clear is skipped *unless* `NotificationListener` has observed a relevant UI-change `AccessibilityEvent` (`TYPE_WINDOW_CONTENT_CHANGED` / `TYPE_WINDOW_STATE_CHANGED` / `TYPE_WINDOWS_CHANGED` / `TYPE_VIEW_SCROLLED`) since the last reset — i.e. event-driven invalidation that mirrors how the framework's own AIC cache invalidates.

- Defaults to `false` → **no behavior change** out of the box.
- **Fails safe**: always clears when the accessibility event listener is inactive (staleness can't be detected without it).
- The explicit `ResetAxCache` endpoint always forces a clear.

## Validation (Android 12 / API 32, real device)
- **Leak (same build, toggle only):** setting OFF → LinearAlloc 4→690 MB (= current baseline); setting ON → flat 3–4 MB across a sustained sleep=0 accessibility-id hammer. simpleperf self-time of `AddMethodToConflictTable` / `art_quick_imt_conflict_trampoline` returns with OFF and is gone with ON.
- **Staleness:** 12 h full-suite run with the fix ON, readback-confirmed across 34 sessions — logged stale-retries 3 vs 2 on the fix-OFF baseline (no spike); no fix-induced stale reads. Remaining failures are pre-existing baseline classes.

## Notes
- A text/locale change emits `TYPE_WINDOW_CONTENT_CHANGED` (in the invalidating set), and no fix-induced staleness was observed; a time-based backstop could be added as belt-and-suspenders if preferred.
- Default-off keeps this safe; happy to propose defaulting it on in a follow-up once it's proven across more app types.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/src/main/java/io/appium/uiautomator2/model/settings/DeferAccessibilityCacheReset.java
+++ b/app/src/main/java/io/appium/uiautomator2/model/settings/DeferAccessibilityCacheReset.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.uiautomator2.model.settings;
+
+/**
+ * When enabled, the server clears the AccessibilityInteractionClient cache before a find/source
+ * lookup only if a relevant UI-change AccessibilityEvent has been observed since the last reset
+ * (rather than on every single lookup). This avoids forcing a full Binder re-traversal of the
+ * accessibility tree per find, which drives unbounded ART LinearAlloc growth under sustained
+ * lookups (see appium/appium-uiautomator2-server#774).
+ *
+ * Defaults to {@code false} to preserve the long-standing always-reset behavior. The optimization
+ * also fails safe to always-reset whenever the accessibility event listener is not active (since
+ * staleness cannot be detected without it).
+ */
+public class DeferAccessibilityCacheReset extends AbstractSetting<Boolean> {
+    private static final String SETTING_NAME = "deferAccessibilityCacheReset";
+    private static final Boolean DEFAULT_VALUE = false;
+    private Boolean value = DEFAULT_VALUE;
+
+    public DeferAccessibilityCacheReset() {
+        super(Boolean.class, SETTING_NAME);
+    }
+
+    @Override
+    public Boolean getValue() {
+        return value;
+    }
+
+    @Override
+    public Boolean getDefaultValue() {
+        return DEFAULT_VALUE;
+    }
+
+    @Override
+    protected void apply(Boolean deferAccessibilityCacheReset) {
+        value = deferAccessibilityCacheReset;
+    }
+}
--- a/app/src/main/java/io/appium/uiautomator2/handler/ResetAxCache.java
+++ b/app/src/main/java/io/appium/uiautomator2/handler/ResetAxCache.java
@@ -30,7 +30,8 @@ public class ResetAxCache extends SafeRequestHandler {
 
     @Override
     protected AppiumResponse safeHandle(IHttpRequest request) {
-        resetAccessibilityCache();
+        // Explicit reset endpoint: always clear, regardless of the deferred-reset optimization.
+        resetAccessibilityCache(true);
         return new AppiumResponse(getSessionId(request));
     }
 }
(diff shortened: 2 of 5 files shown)


app/src/main/java/io/appium/uiautomator2/model/settings/DeferAccessibilityCacheReset.java
```diff
@@ -0,0 +1,53 @@
...
+ * lookup only if a relevant UI-change AccessibilityEvent has been observed since the last reset
+ * (rather than on every single lookup). This avoids forcing a full Binder re-traversal of the
+ * accessibility tree per find, which drives unbounded ART LinearAlloc growth under sustained
+ * lookups (see appium/appium-uiautomator2-server#774).
+ *
+ * Defaults to {@code false} to preserve the long-standing always-reset behavior. The optimization
+ * also fails safe to always-reset whenever the accessibility event listener is not active (since
+ * staleness cannot be detected without it).
+ */
+public class DeferAccessibilityCacheReset extends AbstractSetting<Boolean> {
```
mykola-mokhnach, line 30: what is a potential danger of releasing this patch without an explicit setting (e.g. enabled by default)?

mykola-mokhnach (approved):

mykola-mokhnach:
To me it looks it might be a bug in the Android Accessibility framework. 
Theoretically `AddMethodToConflictTable()` should only be called once. Could it be internally Android creates lots of new classes without reusing existing ones?

@philip-hh Please report this behaviour as a bug to Google with detailed description and steps to reproduce, and lets see what their engineers reply.

KazuCocoa (approved):

app/src/main/java/io/appium/uiautomator2/model/settings/DeferAccessibilityCacheReset.java
philip-hh: The main danger of enabling it by default is **stale reads**. Deferring the per-find `clearAccessibilityCache()` relies on `AccessibilityEvent`s to know the UI changed — `TYPE_WINDOW_CONTENT_CHANGED` / `TYPE_WINDOW_STATE_CHANGED` / `TYPE_WINDOWS_CHANGED` / `TYPE_VIEW_SCROLLED`. If a UI mutation doesn't emit one of those (or the event is coalesced/dropped/delayed relative to the lookup), a `findElement` could match against stale nodes that the unconditional per-find reset would have refreshed. That's a correctness regression that would affect every app and OS version and be hard to attribute — which is why I gated it behind an opt-in default that preserves today's behavior.

In practice it held up well: a 12 h full-suite run with the setting on, readback-confirmed across 34 sessions, showed stale-retries in the low single digits, in line with our fix-off runs (no spike), and — in our validation — no fix-induced stale reads on a genuinely changing UI. So I'd be comfortable defaulting it on after broader validation across more apps/OS versions — and to make default-on safe I'd add a **time-based backstop** (force a reset at least every N ms) so a missed event can't pin stale state indefinitely, and/or widen the invalidating-event set. Happy to either (a) keep it opt-in for this PR and flip the default in a follow-up once it's proven more broadly, or (b) add the backstop and enable by default here — your call.

philip-hh:
> Please report this behaviour as a bug to Google

Agreed it looks like an Android framework/ART issue, and I'll file it with Google.

One nuance for accuracy: it isn't new classes being created — `-verbose:class` shows **zero** new class loads during the leak. It's `ClassLinker::AddMethodToConflictTable` re-allocating the per-class **IMT conflict table** in `LinearAlloc` (→ `MemMapArenaPool::AllocArena` → `MemMap::MapAnonymous`) and apparently not reusing the prior one, so the metadata accumulates for an already-loaded, fixed class set without any class loading. As you said, that table should normally be built once — so the open question for Google is whether it's being rebuilt redundantly rather than reused under sustained interface-method dispatch.

I'll file it against the Android framework/ART (API 31/32) with the simpleperf attribution and a generic repro, and link the tracker issue back here.

philip-hh:
@mykola-mokhnach filed with Google as requested: https://issuetracker.google.com/issues/529701656 (under ART, with the simpleperf attribution and a generic repro). I'll relay whatever their engineers come back with.

KazuCocoa:
Hi @philip-hh , congrats: the Appium project wants to compensate you for this contribution! Please reply to this comment mentioning @jlipps and @KazuCocoa and sharing your [OpenCollective](https://opencollective.com/) account name, so that we can initiate payment! Or let us know if you decline to receive compensation via OpenCollective. Thank you!

PR #789, opened 2026-07-28
chore: Integrate oxc and release configs

mykola-mokhnach:


package.json
```diff
@@ -2,49 +2,47 @@
...
+  },
   "scripts": {
     "bump-gradle-version": "node ./scripts/gradle-version-update.mjs --package-version=${npm_package_version} && git add gradle.properties",
     "build-vendor-jar": "./gradlew :vendor-xpath2:jar",
     "build": "npm run build-vendor-jar && ./gradlew clean assembleServerDebug assembleServerDebugAndroidTest && npm run move-apks && npm run sign-apk",
     "sign-apk": "node ./scripts/sign-apk.mjs",
     "lint:java": "./gradlew lint",
-    "lint": "eslint .",
+    "lint": "oxlint -c oxlint.config.mjs .",
     "lint:fix": "npm run lint -- --fix",
```
eglitise: Should this script also be changed to match other similar PRs?

mykola-mokhnach: both variants work exactly the same. Would you prefer to have `oxlint -c oxlint.config.mjs --fix .` instead?

eglitise: It would simply be consistent, I guess

eglitise (approved):

package.json
mykola-mokhnach: updated

